PR #43, opened 2015-09-17
Feature: User-defined dataset driven tests

Snaipe:
From the discussion at https://github.com/Snaipe/Criterion/issues/40, a use-case has come to light:

>  [...] to implement parameterized tests, that would iterate over a user-generated dataset rather than the cartesian product of many datasets.

Also, this will run a separate test for each member of dataset and hence yield better result stats.

e.g. current approach is to enumerate over `TheoryDataSet` placeholder data array, which results in:

>
[====] Synthesis: Tested: 1 | Passing: 0 | Failing: 1 | Crashing: 0

for 1276 data items tested out of which (for example) only 26 are failing. 

With `ParameterizedTest` feature, we expect to get:

>
[====] Synthesis: Tested: 1276 | Passing: 1250 | Failing: 26 | Crashing: 0

Issue #40, opened 2015-09-14
Feature Request - As a theorist, I need the TheoryDataPoint context in ASSERT report hook


am11:
Hello,

In our current effort of switching from Ruby-based test runner to Criterion, I have encountered a scenario where we are alerting the observer about a TODO test, which unexpectedly passes:
https://github.com/sass/sass-spec/blob/6d3835/lib/sass_spec/test.rb#L26 &
https://github.com/sass/sass-spec/blob/6d3835/lib/sass_spec/test.rb#L37

In order to replicate that effect with Criterion, we have experimented [with ASSERT report hook](https://github.com/am11/libsass/blob/2435ace5e/test/spec_theories.c#L142) which although emits the status of assertion `test->passed`, but does not provide context for Theory's arguments or expected/actual result. Analogous to `struct criterion_theory_stats` for theory, it would be of great help if the existing `struct criterion_assert_stats` or a new `struct criterion_theory_assert_stats` let us exercise:

```c
// Supposedly, it requires a separate `THEORY_ASSERT`
ReportHook(THEORY_ASSERT)(struct criterion_theory_assert_stats *test) {
  if(test->passed && should_test_todo && strstr(test->formatted_args, "-todo-")) {
    // alert unexpectedness 
  }
}
```
---
Nitpick:
On a related note, it seems like `test->message` always contains the error string, even when `test->passed` is `true`.

Snaipe:
The issue with exposing the data points to a generic report hook is that it would effectively be quite error prone; you'd have a (size_t, void*) tuple that you would need to correctly manage on the report hook end, because the parameter tuple is theory-specific while the hook runs at a global scope.

From what I gather in what you want to do, is that you would like to expect a test to fail. That feature is currently lacking, but can be added without any trouble.

Effectively, you'd then have two theories:

```c
void test_for_input(const char *input_path);

Theory((const char *input_path), my_suite, todo_tests, .expect_fail = 1) {
    cr_assume(strstr(input_path, "-todo-"));
    test_for_input(input_path);
}

Theory((const char *input_path), my_suite, non_todo_tests) {
    cr_assume(!strstr(input_path, "-todo-"));
    test_for_input(input_path);
}
```

(assuming a common dataset).

Would that be better ?

am11:
I see this is an interesting approach. However, I think I have missed some info in the first post:

[...]

PR #44, opened 2015-09-17
Add parameterized tests

Snaipe:
From the original issue (#43):

From the discussion at https://github.com/Snaipe/Criterion/issues/40, a use-case has come to light:

>  [...] to implement parameterized tests, that would iterate over a user-generated dataset rather than the cartesian product of many datasets.

Also, this will run a separate test for each member of dataset and hence yield better result stats.

e.g. current approach is to enumerate over `TheoryDataSet` placeholder data array, which results in:

>
[====] Synthesis: Tested: 1 | Passing: 0 | Failing: 1 | Crashing: 0

for 1276 data items tested out of which (for example) only 26 are failing. 

With `ParameterizedTest` feature, we expect to get:

>
[====] Synthesis: Tested: 1276 | Passing: 1250 | Failing: 26 | Crashing: 0

### DIFF
--- a/src/compat/process.c
+++ b/src/compat/process.c
@@ -290,7 +290,17 @@ s_proc_handle *fork_process() {
     ResumeThread(info.hThread);
     CloseHandle(info.hThread);
 
-    while (!ctx->resumed); // wait until the child has initialized itself
+    // wait until the child has initialized itself
+    while (!ctx->resumed) {
+        DWORD exit;
+        GetExitCodeProcess(info.hProcess, &exit);
+        if (exit != STILL_ACTIVE) {
+            UnmapViewOfFile(ctx);
+            CloseHandle(sharedMem);
+            CloseHandle(info.hProcess);
+            return (void *) -1;
+        }
+    }
 
     UnmapViewOfFile(ctx);
     CloseHandle(sharedMem);
--- a/src/compat/process.c
+++ b/src/compat/process.c
@@ -183,18 +183,20 @@ int resume_child(void) {
            0,
            MAPPING_SIZE);
 
-    if (ctx == NULL)
+    if (ctx == NULL) {
+        CloseHandle(sharedMem);
         exit(-1);
+    }
 
     local_ctx = *ctx;
     struct test_single_param *param = NULL;
     if (local_ctx.param.size != 0) {
-        param = malloc(sizeof (struct test_single_param) + local_ctx.param->size);
+        param = malloc(sizeof (struct test_single_param) + local_ctx.param.size);
         *param = (struct test_single_param) {
-            .size = local_ctx.param->size,
+            .size = local_ctx.param.size,
             .ptr = param + 1,
         };
-        memcpy(param + 1, local_ctx.param->ptr, param->size);
+        memcpy(param + 1, local_ctx.param.ptr, param->size);
     }
 
     g_worker_context = (struct worker_context) {
@@ -282,6 +284,7 @@ s_proc_handle *fork_process() {
     if (g_worker_context.param) {
         ctx->param = *g_worker_context.param,
         memcpy(ctx + 1, g_worker_context.param->ptr, g_worker_context.param->size);
+        ctx->param.ptr = ctx + 1;
     }
 
     if (g_worker_context.suite->data)
@@ -293,7 +296,9 @@ s_proc_handle *fork_process() {
     // wait until the child has initialized itself
     while (!ctx->resumed) {
         DWORD exit;
-        GetExitCodeProcess(info.hProcess, &exit);
+        if (!GetExitCodeProcess(info.hProcess, &exit));
+            continue;
+
         if (exit != STILL_ACTIVE) {
             UnmapViewOfFile(ctx);
             CloseHandle(sharedMem);
--- a/src/compat/process.c
+++ b/src/compat/process.c
@@ -100,10 +100,10 @@ struct full_context {
     struct pipe_handle pipe;
     struct test_single_param param;
     HANDLE sync;
+    DWORD extra_size;
 };
 
 static TCHAR g_mapping_name[] = TEXT("WinCriterionWorker_%lu");
-#define MAPPING_SIZE 1024
 
 static struct full_context local_ctx;
 #endif
@@ -183,7 +183,7 @@ int resume_child(void) {
            FILE_MAP_ALL_ACCESS,
            0,
            0,
-           MAPPING_SIZE);
+           sizeof (struct full_context));
 
     if (ctx == NULL) {
         CloseHandle(sharedMem);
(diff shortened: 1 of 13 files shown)


src/compat/process.c
```diff
@@ -186,11 +187,22 @@ int resume_child(void) {
...
 
     local_ctx = *ctx;
+    struct test_single_param *param = NULL;
+    if (local_ctx.param.size != 0) {
+        param = malloc(sizeof (struct test_single_param) + local_ctx.param->size);
+        *param = (struct test_single_param) {
+            .size = local_ctx.param->size,
+            .ptr = param + 1,
+        };
+        memcpy(param + 1, local_ctx.param->ptr, param->size);
```
am11: Seems like these should be `local_ctx.param.ptr`, `local_ctx.param.size` etc. according to MinGW and MSVC. Not sure how it is building with GCC! strange. :smile:

am11:
Bundle of thanks @Snaipe! :+1: :tada:

Snaipe:
Parameterized tests are mostly working -- except on windows.

The test sample hangs/crashes at random, and I can't get any debugger to work since the issue is located in the child process. I suspect there is a memory corruption or array overrun in the fork() code for windows, but I can't be sure without a tool like valgrind.

Snaipe:
After a careful review of the forking code, I spotted a few errors in the context passing from the parent to the child -- parameterized tests should now work seamlessly.

The only thing left is extensive testing, setting up the regression tests for the feature, and removing the hard limit of 1024 bytes of transferred user data in the windows forking code.

Snaipe:
There seem to be a very unfortunate limitation on parameterized tests on windows; you cannot pass parameters containing pointers to dynamically allocated memory. This is because the forking process on windows has no mean to cleanly copy the heap without causing crashes, so everything allocated in the parent process will stay in the parent process; and unlike theories, parameterized tests initialize their parameters in the parent process.

What you *can* do however is dynamically allocate the parameter array; a workaround to the initial limitation would be to explicitely not put any pointer to dynamically allocated memory inside the parameter structure and coalesce everything in it.

Effectively, instead of this:

```c
struct param_tuple {
    struct some_struct *s;
    char *some_string;
};
```

You'd have this:

```c
struct param_tuple {
    struct some_struct s;
    char some_string[1024]; // completely arbitrary, unfortunately
};
```

The only exceptions where one could keep a pointer, is of course for statically allocated data, such as:
* string literals
* pointers to constant global variables
* pointers to constant static local variables

Snaipe:
I'm going to clean this up and merge with the current known limitations on windows -- no work is really left here, and it will make testing easier on the `features/winfork` branch.

am11:
@Snaipe, I am able to build `features/parameterized-tests` branch locally on Windows 10, then I had to copy `.\Release\criterion.dll` to `C:\Program Files (x86)\CMake\bin` to successfully run `ctest` (took me a while to figure this out; is there  a better way?). 
In the end, it is failing one last test (`criterion_unit_tests`):

```powershell
      Start 31: criterion_unit_tests
31/31 Test #31: criterion_unit_tests .............***Failed    0.11 sec

97% tests passed, 1 tests failed out of 31

Total Test time (real) =   4.48 sec

The following tests FAILED:
         31 - criterion_unit_tests (Failed)
Errors while running CTest
```

Snaipe:
> I had to copy .\Release\criterion.dll to C:\Program Files (x86)\CMake\bin to successfully run ctest (took me a while to figure this out; is there a better way?).

Either copy the dll into the test output binary directory (should be `.\test\Release` in your case) or add the output binary directory (i.e. `.\Release`) to your path.

For the failure, I'll investigate once I get my win10 dev environment running.

am11:
I copied `criterion.dll` in `.\Criterion\test\Release` and ran `ctest` from `.\Criterion`, but still the same result. However, adding to PATH (`SET PATH=%PATH%;C:\Users\Adeel\Source\Repos\Criterion\Release\;` requires absolute path..) worked! :)

I am configuring Windows 7 with VS2013 in VM for testing related build stuff.

am11:
FWIW, these prepared images (for web testers) come handy:https://dev.modern.ie/tools/vms/windows/ (just need to install VS community edition after launching the OS).

Snaipe:
I'm actually using [vagrant](https://www.vagrantup.com/) to download an manage the images, it's quite handy.

am11:
Yes it is very handy. I have vagrant installed too, but the sad part is, most of the images do not have Hyper-V provider (e.g. I have Ubuntu 14 setup with vagrant..). Something went wrong with my connections when i installed VirtualBox to make other images work. So meanwhile I am working with raw VMs for FreeBSD and Windows 7 to quickly get to the most exciting point: `compilation`. :)

Snaipe:
I cannot reproduce the failure on my Win10 VM, what's the output of manually running the unit test ?

am11:
```cmd
:: in cmd
C:\Users\Adeel\Source\Repos\Criterion>test\Release\criterion_unit_tests.exe

> [----] C:\Users\Adeel\Source\Repos\Criterion\test\redirect.cc:88: Assertion failed: The expression (read) == ("Foo") is
false.
[FAIL] redirect::stdin_: (0.00s)
[====] Synthesis: Tested: 8 | Passing: 7 | Failing: 1 | Crashing: 0
```

Snaipe:
So apparently, by running the tests locally, I have spotted a bug (#50) that makes test from only one compilation unit run on very rare conditions (that I somehow triggered with my Win10 setup).

I'll see what the redirection issue is about.

am11:
Thanks! Please let me know if you need some debug info that I could pass along (in case you can reproduce redirection assertion failure). :)

Snaipe:
@am11 it seems that the redirection issue is an old one, and the recent fix to #50 just brought that up on the latest bleeding CI builds (this means that it's not specific to features/parameterized-tests).

There's probably no need to open a separate issue for that since it has been living in bleeding only.

Snaipe:
Nevermind that, it seems that the way Microsoft implemented its `std::[io]fstream` is absolutely moronic: doing `stream << "Foo"` actual writes the *pointer* to "Foo" into the stream rather than the string literal *Foo*. I have no idea why this is happening, but in the meantime I have changed "Foo" to std::string("Foo"), and it fixed the unit test.

I'll open an enhancement issue to see what can be done to get `stream << "Foo"` to work.

am11:
@Snaipe, took a closer look at `redirect.c#L88`, maybe we can use `std::cin.rdbuf()` there to redirect stream instead?

Snaipe:
@am11 the redirection works fine, it's just that writing a string literal does not do what you would expect.

am11:
Great! I just read your previous comment. I posted simultaneously. 
Tested with current master and the result is :100:% :+1: 
For the record, I am using the following script in PowerShell to build and run Criterion tests:

```powershell
rm -r -fo s-criterion; git clone https://github.com/Snaipe/Criterion --recursive s-criterion;
cd s-criterion; $env:Path += ";$(pwd)\Release"; cmake .;cmake --build . --config Release;
cmake --build . --target criterion_tests --config Release; ctest
```

PR #47, opened 2015-09-18
Reimplement fork() for windows

Snaipe:
The current fork() implementation for windows is causing unfortunate effects on parameterized tests (#44).

Inspection of the internal NT 6.0 API functions shows interesting functions such as RtlCloneUserProcess -- maybe this will help.

Otherwise, it would be preferable to try and copy the heap from the parent to the child, or define some context passing rules (maybe implement our own tiny windows-specific memory allocator that can be inherited?)


Snaipe:
Uh oh.

I'm currently downloading a windows 10 VM to be sure, but I think that `RtlCloneUserProcess` might have been removed in more recent versions of `ntdll.dll`.

This would mean that doing a clean fork() wouldn't be worth the maintainance and overall trouble -- it has been Cygwin's problem for years, and I sure don't want it to be mine, too.

If things get dark and I can't get the fork() to work, I will allocate a fixed region on the virtual address space and make it inherited by the child process. This could be a good tradeoff.

am11:
AFAICT, `ntdll.h` wasn't present on Windows 7 and 8x SDKs either? It is certainly not available on Windows 10. Was it ever the official header by Microsoft (as I can't find it on MSDN library)?

Snaipe:
`ntdll.h` does not exist, these are internal, and hence undocumented, functions of the NT kernel API. I have to dynamically hook onto ntdll.dll to access those functions.

Snaipe:
I have confirmed that RtlCloneUserProcess has been removed since around NT 6.2 -- so this code will only work on Windows 7, Vista, and Server 2008.

I'm going to revert the changes and go with another idea: implement `cr_malloc`, `cr_free`, `cr_calloc` and `cr_realloc`.

On *nix systems, these four identifiers will be macros defined to malloc, free, calloc and realloc respectively -- however, on windows, it will actually call a function that will manipulate a private heap created by `HeapCreate`.

`fork_process()` will then proceed to walk the heap and copy any allocated region into the child.

This would be potentially unsafe for the child, unless we explicitely restrict the usage of `cr_*` allocation functions: if we define their usage to be exclusive to functions defined by `ParameterizedTestParameters`, then the heap will not be manipulated further in the child process.

Snaipe:
The appveyor build seems to hang at random when executing the tests, and I cannot reproduce this locally.
I'll take a shot at using the virtual desktop feature of appveyor to see what's wrong.

am11:
I added this line in `appveyor.yml`:

```yaml
init:
- ps: iex ((new-object net.webclient).DownloadString('https://raw.githubusercontent.com/appveyor/ci/master/scripts/enable-rdp.ps1'))
```

to [enable remote desktop](http://www.appveyor.com/docs/how-to/rdp-to-build-worker), and it ran all the tests with 100% success.
https://ci.appveyor.com/project/am11/criterion/build/2.0.1_b1-bleeding

For some reason, without RDP config, it is failing on AppVeyor.. (&#3232;_&#3232;)

I have found some related links, perhaps which may come handy:
* https://msdn.microsoft.com/en-us/library/kdzttdcb.aspx
* https://msdn.microsoft.com/en-us/library/aa365574.aspx (under "Using Pipes for IPC" headline)

Snaipe:
Well then.

HeapCreate works fine. However, the failures reported by appveyor happen completely randomly. The child process crashes right before it hits main(), and it happens at random with a 1/60-ish chance. I think this is due to the fact that the second heap sometimes overlaps with a region needed by the process initialization, but not yet commited since the process is created suspended.

It might as well be the flying spaghetti monster meddling with the process memory with its noodly appendage, so I'm kinda split between reimplementing my own allocator completely (kinda overkill, but this is forcing my hand) or just calling fork_process() again if it fails once, since failure twice in a row would be very unlikely.

am11:
Just a random thought; would tapping into vtable and playing with [adjustor thunk](http://blogs.msdn.com/b/oldnewthing/archive/2004/02/06/68695.aspx) worth a trouble for this task?

Snaipe:
An adjustor thunk acts just as a trampoline for a virtual method call; here the issue is a bit different, since it's probably about the ASLR picking a slide value that collides with the previously allocated private heap.

am11:
https://github.com/am11/Criterion/commit/40f5a9fa5b1b522644575227e4e58ae43aeb92a9.

It was hanging with debug configuration. With release configuration, it passes all the test with MSVC: https://ci.appveyor.com/project/am11/criterion/build/2.0.2_b18-features/winfork/job/hj5a023s9s893690 but still hanging with MinGW.  I reordered the matrix to make MSVC build run first. :)

am11:
I ran `ctest` around 30 times and able to capture two call stacks: https://gist.github.com/am11/4c8f0d312a242bdc5357.

Snaipe:
Yeah, that's what I though, the private heap is conflicting with the default heap during initialization. You would think that the pre-main section would try to find an empty space instead of immediately giving up...

Snaipe:
Okay, there we go. Seems that rolling our own allocator works just fine.

Now the rest of the work is to get rid of the temporary piece of garbage allocator that 9ef2d30 introduced and use an usable implementation.

Unless I find a lightweight windows-only allocator library that can be easily used, I'm probably going to write it myself.

am11:
:+1: 
I have found one https://github.com/emeryberger/DieHard via https://github.com/search?utf8=%E2%9C%93&q=windows+allocator :)

am11:
This is still giving me 404: https://github.com/microsoft/windows .. otherwise, I would have opened a Pull Request or a Feature Request to fix `Create[Process/Thread]`.. :stuck_out_tongue_winking_eye:

Snaipe:
It seems that most of the memory allocators I find does not really fit what we need:

* Most implementations offer no mean to walk the heap
* They are not designed to be inherited from a parent process to a child process (no copy on write)

So given that I'm probably going to fork jemalloc or other and specialize it for our purposes.

am11:
Do either of these useful:

* [_beginthread, _beginthreadex](https://msdn.microsoft.com/en-us/library/kdzttdcb.aspx)
* [_CrtIsValidHeapPointer](https://msdn.microsoft.com/en-us/library/ys6cfhhh.aspx) 
* [_get_heap_handle]()
* [_heapadd](https://msdn.microsoft.com/en-us/library/wykz6chs.aspx)
* [_heapchk](https://msdn.microsoft.com/en-us/library/2xse74he.aspx)
* [_heapmin](https://msdn.microsoft.com/en-us/library/fc7etheh.aspx)
* [_heapset](https://msdn.microsoft.com/en-us/library/y6118zwe.aspx)
* [_heapwalk](https://msdn.microsoft.com/en-us/library/h0c183dk.aspx)

Snaipe:
_beginthread(ex) creates a thread, not a process -- it cannot be applied to our use case.

All other functions are related to HeapAlloc, and HeapAlloc fails randomly according to the recent builds.

I'll try to see if I can somehow 'hint' to the process to allocate the heap in higher addresses, but I don't have much hope.

Snaipe:
Nevermind that.

I have a quick'n'dirty fix that involves recreating a heap over and over until the base address is out of the "death zone". It seems to work without really slowing down the test runner, so I think I'll roll with that.

PR #39, opened 2015-09-22
Make the tests run in parallel

Snaipe:


Snaipe:
Parallelization is working very well, the only thing left is to clean up obsolete code and add a command line switch + environment variable.

am11:
Excellent! :tada: 
Does this parallelism / asynchrony work with all types of tests?

Snaipe:
@am11 Yes, except for theory iterations (the test running the theory is run in parallel with all the other test, but the content of the theory is not parallelized).

PR #41, opened 2015-09-25
Float eq/neq assertions are not accurate

Snaipe:
Two adjascent floats can be considered not equal if the choice of epsilon is unwise. This becomes problematic because the choice of epsilon depends on the order of the two floating point values one is comparing; for instance, between 1e10 and 1e11, an epsilon of 1 is acceptable, while it is not for 1 and 0.9.

Instead, we should be comparing the ULPs of the difference between the two values, effectively using the N next adjascent floats as upper and lower bound for equality.
It should also be noted that FLOAT_MAX and INFINITY are within 1 ULP from each other, so these special cases should also be handled.

Since we are comparing arbitrary floating-point values, there is seemingly no silver bullet, as ULPs are effective for nonzero value comparison, while epsilons are.
Great care should be used while implementing a solution that satisfies any scenario.

An article describing the technique can be found [here](https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/)

### DIFF
--- a/src/float/ieee754.c
+++ b/src/float/ieee754.c
@@ -135,6 +135,44 @@ IMPL_CMP_RELEPS(double)
 IMPL_CMP_ULP(single)
 IMPL_CMP_ULP(double)
 
+int ieee754_single_cmp_adaptive(register union criterion_fp_single val1,
+                                register union criterion_fp_single val2) {
+
+    size_t max_ulps = fp_range(single);
+
+    const int p1 = fpclassify(val1.f),
+              p2 = fpclassify(val2.f);
+
+    if (unlikely((val1.i == val2.i
+                    || (p1 == FP_ZERO && p2 == FP_ZERO)
+                    || ((p1 == FP_ZERO) ^ (p2 == FP_SUBNORMAL))
+                    || ((p2 == FP_ZERO) ^ (p1 == FP_SUBNORMAL)))
+            && p1 != FP_NAN && p2 != FP_NAN))
+        return 0;
+
+    if (unlikely(p1 == FP_INFINITE || p2 == FP_INFINITE))
+        return val1.f < val2.f ? -1 : 1;
+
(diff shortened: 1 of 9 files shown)


src/float/ieee754.c
```diff
@@ -0,0 +1,159 @@
...
+    return x < 0 ? -x : x;
+}
+
+#ifdef __GNUC__
+# define likely(x)   __builtin_expect((x),1)
+# define unlikely(x) __builtin_expect((x),0)
+#else
+# define likely(x)
+# define unlikely(x)
+#endif
```
am11: Changing this `ifdef` block to:

```C
#ifndef __GNUC__
# define __builtin_expect(exp, c) (exp)
#endif
# define likely(x)   __builtin_expect((x),1)
# define unlikely(x) __builtin_expect((x),0)
```

builds on VS2015.

Snaipe:
This will be on hold until after the next release. Implementing the comparison functions is non-trivial, and I need to research good defaults for each strategy before hoping to merge anything back to bleeding.

Backward compatibility also needs to be addressed for the ``cr_assert_float_*`` assertion family.

test/float.cc
```diff
@@ -0,0 +1,69 @@
+#include "criterion/parameterized.h"
+#include "criterion/options.h"
+#include "criterion/theories.h"
+#include "criterion/float.h"
+#include <cmath>
+#include <cfloat>
+#include <iostream>
+#include <unistd.h>
```
am11: This can probably be changed to:

```c
#include <iostream>

#ifdef _MSC_VER
# include <io.h>
#else
# include <unistd.h>
#endif
```

Snaipe: Actually unistd.h isn't used at all, it's just a forgotten artifact that I'll most probably remove later.

Snaipe:
Closing this since #182 provides all of what this PR promises, but better.

PR #71, opened 2015-10-04
Support multiple languages

Snaipe:
Criterion has the interesting property of having the tests and the runner only linked by the ABI. This means that, unlike a lot of frameworks, tests are in no way dependent on the runner code, nor features of the C language; hence Criterion could run tests from any language that compiles to native (C, C++, Objective C, D, Rust, Haskell, OCaml, ...)

This PR does not plan to support all of these languages, but does provide the necessary changes in the framework to allow anyone to simply write a wrapper for the desired language.

It additionally adds support for Objective C on *nixes as a proof of concept, since the language is nothing more than a superset of the C language, and is supported by both GCC and Clang.


Snaipe:
This is currently blocked by travis-ci/apt-package-whitelist#1363.

PR #73, opened 2015-11-07
export test results to JSON

Snaipe:
XML sucks, we want JSON :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -135,6 +135,7 @@ set(SOURCE_FILES
   src/io/output.h
   src/io/tap.c
   src/io/xml.c
+  src/io/json.c
   src/log/logging.c
   src/log/normal.c
   src/string/i18n.c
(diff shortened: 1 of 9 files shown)


Snaipe:
The feature is implemented; the only things left are, in the scope of this PR:
* Document both `--json` and `CRITERION_ENABLE_JSON`
* Add tests & regression tests on the json outputs

am11:
Thanks for this @Snaipe! :+1: 
In order to make it even more extendable, would it make sense to have "custom reporter" functionality, so the consumer can provide function pointer to their reporter (say if they rather want it in YAML or CSON format or they want XML report written differently) and Criterion just call it back with a single well-defined `struct criterion_results` object instead? This approach is used by some node.js based linters (CofeeLint, CppLint, JSHint, JSCS, tslint and since so forth).

Snaipe:
@am11 This is already almost-ish done by the current logger structure; someone can easily process the results of all the tests, if they follow a structure like the one used by the TAP, XML or Json reporter.

However, the system itself is lacking, and issue #68 is discussing what is lacking right now. Namely, the idea would be to decouple this reporting from the logging system, and allow users to register more easily custom (and multiple) reporters.

PR #84, opened 2015-12-03
don't print code coverage for gcov

D3Hunter:
I have test 2.1.0 and 2.1.1, still have this problem on Unit-test programs


Snaipe:
You need to either pass ``--no-early-exit`` or define CRITERION_NO_EARLY_EXIT=1 in your environment to get proper coverage.

I guess this should be mentioned more in the FAQ, will add it.

PR #86, opened 2015-12-08
Decouple the multiple language support from the test runner

Snaipe:
The scope of this PR is to remove all the language wrappers in criterion and let the compilation of language-specific wrappers happen in user code, i.e. by defining the wrappers in the internal API headers and using them as test entry points.

This has the benefit of decoupling the test logic from language specifics, and sever any unneeded dependencies brought by external languages: you only get what you pay for.

This was immediately visible when compiling with Objective-C support: Criterion would have a hard dependency on libobjc.so AND the GNUstep framework, which smells just like bad decisions for something that was supposed to be a proof of concept on the support of multiple languages.

Also, this means that the C++ support can be maintained *without* an actual C++ compiler on the build server.

Another benefit of doing this decoupling is that after the merge, the only layer between tests and the runner will be the ABI, and more precisely the cdecl calling convention of a void-returning-void function, making bridges for native languages much simpler to write.


am11:
:+1:

* Is it the same mechanism as the foreign function interface ([FFI](https://en.wikipedia.org/wiki/Foreign_function_interface))?
  * Does Criterion provide its own form of FFI so other langs can interact?
  * Or is it the reciprocal; the kind of interface which engages other langs' FFI support?
* Would it make sense to add examples for some languages under `examples/` directory?

Snaipe:
Currently, the interface between tests in this PR will be the cdecl calling convention, so the target language must have some kind of FFI layer supporting the default calling convention, which most probably does if the language compiles to native.

Ideally, we would simply prepare a new stack frame and manually jump on the specified address, eliminating the calling convention altogether -- however, the language would have to support naked functions, and this is less likely to happen than a FFI layer.

Issue #69, opened 2015-10-02
Provide a minimal no-dependency server/client system for embedded devices


Snaipe:
Embedded systems are not guaranteed to provide a POSIX environment or have the hardware requirements to even run the full runner.

It could be a nice addition for embedded developers to compile the hardware tests in "client" mode. 
The test runner would then run on some kind of "server" mode that listens for any data stream from the client.

Food for thought:

* Should the server start the client test itself, by asking a user-defined function that handles the logic of launching the test, or should it just wait for the client to connect by itself ? (maybe add both options and let the user decide)
* Should the server try to upload the client source or client binary to the device, using an user defined function, or should it assume that the device has the client and is already ready for testing?
* How should parameterized tests or theories be handled on hardware tests ?

On thing that is clear is that most of the handling logic will have to be provided by the user himself. This includes streaming data between client and server, compiling the client and programming the chip with it, running the client, ...

kylemanna:
My dream would be to have a test framework that allows for CI and running of test suites on real world ARM Cortex-M1/M3/M4 devices.  CI on real hardware after every pull request, merge, etc.

A CI tool would cross compile the code and then run [openocd](http://openocd.org/) or similar to flash the device with the latest firmware build with some interface to communicate with the "server" runner.  The device under test should reset after programming and be waiting for communications via a well defined communications stream.  As long as a bidirectional stream is possible, any transport could be implemented such as BLE, UART, ARM semihosting, TCP/IP.

The server process on the host PC would then start the test on the target embedded processor.  This may require a short arbitrary message is sent to the embedded device over a UART (or similar) and then enter the client-server communications interface.

Additionally, embeddeding the tests completely in the firmware would allow a minimal version of criterion to serve as a POST for the device.  In this manner, invoking the test suite by a function call would be great idea.

An interesting test framework that targets embedded devices is https://cmocka.org/

PR #89, opened 2015-12-09
Rewrite the I/O logic and define a protocol

Snaipe:
This is mostly here to discuss and test anything related to the new protocol definition and the rewrite of the I/O logic.

The scope of this PR is to implement a client/server model, to start implementing the feature request #69.

The current technologies under test are:
* Protocol definition using the protobuf DSL.
* Protobuf implementation provided by nanopb.
* I/O abstraction provided by nanomsg.

nanopb should be more adequate than libprotobuf for serialization, as there seems to be a lot of effort done to make in run on embedded systems.


Snaipe:
The current protocol definition can be found [here](https://github.com/Snaipe/Criterion/blob/features/io-rewrite-nanopb/src/protocol/criterion.proto)

ckaran:
Looks good!  Although if you're going with protocol buffers, do you need the version field?  Protocol buffers support a method of updating messages: https://developers.google.com/protocol-buffers/docs/proto3#updating.  That said, I don't know what [nanopb](http://koti.kapsi.fi/jpa/nanopb/) supports in terms of updated messages, so it may still be necessary.

Snaipe:
The version field is mostly here to allow updated versions of the protocol that would be otherwise breaking backward compatibility -- although it might never be used at all, its just a safeguard. Message updating will still be used within a protocol version.

ckaran:
OK, but isn't that the point of protocol buffers?  So that you can detect if you have backwards-incompatible changes and deal with them in a sane manner?  Or maybe I'm wrong... Eh, you're probably right, even if protocol buffers are able to deal with it, as human beings we like to see what version we're dealing with.  I withdraw the comment! :-)

Snaipe:
If you have backward-incompatible changes with protobuf, you're kinda screwed because the implementation deserialize the message as the new protocol regardless. This is why they tell you to never do that by never reusing field numbers (or making them reserved if you remove some), and not adding new required fields.

Snaipe:
The usage of nanomsg for the purpose of this rewrite might be compromised. I'm waiting for confirmation on my fears from the developers themselves, but here's the gist of it:

* nanomsg internally heavily uses threads and mutexes
* they don't register a pthread_atfork handler to release all associated mutexes and put the library in a 'forkable' state.

This means that nanomsg is effectively unusable after a fork since all mutexes are inherited by the fork, and the child might have resumed when one of the threads was in a critical section.

This issue, coupled to the inheritance semantics of file handles on windows, makes it excruciatingly hard to find a proper I/O library that fits our need.

Hence, I'm going to try two things:

1. Discuss this further with the developers of nanomsg in the hope that we can find a reasonable workaround for this issue,
2. Or, if the above attempt definitely isn't leading anywhere, implement my own I/O library for the express purpose of being fork()'able and reliable on all platforms.

And now, we wait.

ckaran:
Man, I really hope that there is a good way to handle fork().  The only thing I've found so far about it is http://www.freelists.org/post/nanomsg/threading-guidelines-and-is-traditional-unix-single-threaded-mode-possible.  It suggests closing all nanomsg sockets before forking, but I'm not sure how IPC is handled afterwards.  Is it possible to open the sockets after the fork?  With luck, things have changed for the positive in the intervening year+...

Snaipe:
I forked nanomsg and started to fix the fork behaviour. A PR reporting the work in progress can be found at nanomsg/nanomsg#587.

Currently, the rewrite is almost done -- the only thing left is extensive testing.

The new protocol defines a much more rich test lifecycle, but I purposely restricted its interpretation on the runner side for backward compatibility purposes -- it will likely grow out of it for the 3.0.0 release.

the ReportHook API will most likely be deprecated and removed for 3.0.0 -- a replacement will be set up, that properly handles nested subtest phases as defined by the protocol.

Snaipe:
I'm going to merge this into bleeding -- the only bugs left come from the WIP on fork support for nanomsg and will be fixed in time. However, as there is a maintainer change undergoing over nanomsg and hence slower processing of PRs and reviews, and because the bleeding branch is slowly diverging from the changes introduced by the rewrite, it becomes necessary to merge the changes and iterate further on the new bleeding branch.

am11:
:+1:

ckaran:
I'm coming to the party late, but :+1:  :)

PR #93, opened 2015-12-29
cmake: add a SOVERSION to libcriterion.so

hppritcha:
Shared libraries should have an SONAME
to facilitate proper dynamic linking.

@Snaipe it would be nice for the criterion shared library to have 
a proper SONAME.  

Signed-off-by: Howard Pritchard <howardp@lanl.gov>


Snaipe:
Fair enough, it seems that the SO version got stripped when we moved from autotools to cmake. Thanks!

PR #100, opened 2016-01-28
Fixed doc/assert: missing size parameter

MrAnno:
Size parameter was missing from `cr_assert_arr_eq` and `cr_assert_arr_neq`.


Snaipe:
Thanks for catching this!

PR #101, opened 2016-02-11
Visibility change:

kaidowei:
The numbers next to Failing and Crashing in the text synthesis will not be printed red, if they are zero.


Snaipe:
I'm going to leave this PR alone for a little while and get some more opinions on the patch, as this change the visual rendering of the tool. I'm not that much of a fan myself to be honest, as this makes the output inconsistent style-wise, but I can respect the ideas behind it.

kugel-:
I support this patch. The red 0 always makes me nervous, even though all tests pass!

PR #102, opened 2016-02-13
added german translation

a1lu:
added german translation at po/de.po
added "de" to po/LINGUAS

LANGUAGE=de_DE ./simple.bin prints german strings


Snaipe:
Merged manually on bleeding rather than master.

PR #106, opened 2016-02-18
Add doxygen documentation in API headers (Fixes #80)

### LINKED ISSUE #80: Add doxygen-like docs
A lot of C and C++ users read the public header files for documentation. It would be a great benefit to add doxygen-like docs on API functions and macros.

* [x] ``criterion/abort.h``
* [x] ``criterion/alloc.h``
* [ ] ``criterion/assert.h``
* [x] ``criterion/criterion.h``
* [ ] ``criterion/event.h``
* [x] ``criterion/hooks.h``
* [ ] ``criterion/logging.h``
* [x] ``criterion/options.h``
* [x] ``criterion/output.h``
* [x] ``criterion/parameterized.h``
* [x] ``criterion/redirect.h``
* [ ] ``criterion/stats.h``
* [x] ``criterion/theories.h``
* [x] ``criterion/types.h``

a1lu:
Added doxygen docs to the API at include/criterion


include/criterion/redirect.h
```diff
@@ -100,42 +104,441 @@ CR_API CR_STDN FILE *cr_mock_file_size(size_t max_size);
...
+ * Passes if the contents of \c stdout are equal to the contents of \c RefFile
+ *
+ * Call: \c cr_assert_stdout_eq(RefFile, [FormatString,
+ * [Args...]])\n
+ * Passes if the contents of \c stdout are equal to the contents of \c RefFile.
+ * Otherwise the test is marked as failure and the execution of the function
+ * is aborted.
+ * The optional string is printed on failure.
+ *
+ * @param[in] ExpectedContents C string with the ExpectedContents
```
Snaipe: I believe this should be `RefFile` rather than `ExpectedContents`.

include/criterion/redirect.h
```diff
@@ -100,42 +104,441 @@ CR_API CR_STDN FILE *cr_mock_file_size(size_t max_size);
...
+/**
+ * Passes if the contents of \c stdout are equal to the contents of \c RefFile
+ *
+ * Call: \c cr_expect_stdout_eq(RefFile, [FormatString,
+ * [Args...]])\n
+ * Passes if the contents of \c stdout are equal to the contents of \c RefFile.
+ * Otherwise the test is marked as failure but the execution will continue.
+ * The optional string is printed on failure.
+ *
+ * @param[in] ExpectedContents C string with the ExpectedContents
```
Snaipe: Same comment for `RefFile`

include/criterion/redirect.h
```diff
@@ -100,42 +104,441 @@ CR_API CR_STDN FILE *cr_mock_file_size(size_t max_size);
...
+ * RefFile
+ *
+ * Call: \c cr_assert_stout_neq(RefFile, [FormatString, [Args...]])\n
+ * Passes if the contents of \c stdout are not equal to the contents of \c
+ * RefFile.
+ * Otherwise the test is marked as failure and the execution of the function
+ * is aborted.
+ * The optional string is printed on failure.
+ *
+ * @param[in] UnexpectedContents C string with the UnexpectedContents
```
Snaipe: Same comment for `RefFile`

include/criterion/redirect.h
```diff
@@ -100,42 +104,441 @@ CR_API CR_STDN FILE *cr_mock_file_size(size_t max_size);
...
+ * Passes if the contents of \c stdout are not equal to the contents of \c
+ * RefFile
+ *
+ * Call: \c cr_expect_stout_neq(RefFile, [FormatString, [Args...]])\n
+ * Passes if the contents of \c stdout are not equal to the contents of \c
+ * RefFile.
+ * Otherwise the test is marked as failure but the execution will continue.
+ * The optional string is printed on failure.
+ *
+ * @param[in] UnexpectedContents C string with the UnexpectedContents
```
Snaipe: Same comment for `RefFile`

include/criterion/assert.h
```diff
@@ -30,109 +35,1041 @@
...
+ */
+
+/**
+ * Fails always.
+ *
+ * Call: \c cr_assert_fail([FormatString, [Args...]]).\n
+ * The test is marked as failure and the execution of the function is aborted.
+ * The optional string is printed on failure.
+ *
+ * @param[in] fmt printf like format string
```
Snaipe: I just realized, but `fmt` should be `FormatString`.

include/criterion/assert.h
```diff
@@ -30,109 +35,1041 @@
...
+ * @param[in] Expected Expected array
+ * @param[in] fmt printf like format string
+ *
+ *****************************************************************************/
 # define cr_assert_arr_eq(...) CR_EXPAND(cr_assert_mem_op_va_(CR_FAIL_ABORT_,      ==, __VA_ARGS__))
+
+/**
+ * Passes if Actual is byte-to-byte equal to Expected
+ *
+ * Call: \c cr_assert_arr_eq(Actual, Expected, [FormatString, [Args...]])\n
```
Snaipe: This should be `cr_expect_arr_eq`

include/criterion/assert.h
```diff
@@ -30,109 +35,1041 @@
...
+ * @param[in] Size Number of bytes to check
+ * @param[in] fmt printf like format string
+ *
+ *****************************************************************************/
 # define cr_assert_arr_neq(...) CR_EXPAND(cr_assert_mem_op_va_(CR_FAIL_ABORT_,     !=, __VA_ARGS__))
+
+/**
+ * Passes if Actual is not byte-to-byte equal to Expected
+ *
+ * Call: \c cr_assert_arr_neq(Actual, Expected, Size, [FormatString,
```
Snaipe: This should be `cr_assert_arr_neq`

include/criterion/assert.h
```diff
@@ -30,109 +35,1041 @@
...
+ * <type of Expected>) to do the comparison
+ * @param[in] fmt printf like format string
+ *
+ *****************************************************************************/
 #  define cr_assert_arr_eq_cmp(...) CR_EXPAND(cr_assert_arr_op_cmp_va_(CR_FAIL_ABORT_,      ==, __VA_ARGS__))
+
+/**
+ * Passes if Actual is comparatively equal to Expected (C++ / GNU C99 only)
+ *
+ * Call: \c cr_assert_arr_eq_cmp(Actual, Expected, Size, Cmp, [FormatString,
```
Snaipe: This should be `cr_assert_arr_eq_cmp`

include/criterion/assert.h
```diff
@@ -30,109 +35,1041 @@
...
+ *****************************************************************************/
 # define cr_assert_arr_eq(...) CR_EXPAND(cr_assert_mem_op_va_(CR_FAIL_ABORT_,      ==, __VA_ARGS__))
+
+/**
+ * Passes if Actual is byte-to-byte equal to Expected
+ *
+ * Call: \c cr_assert_arr_eq(Actual, Expected, [FormatString, [Args...]])\n
+ * Passes if Actual is byte-to-byte equal to Expected.
+ * Otherwise the test is marked as failure but the execution will continue.
+ * The optional string is printed on failure.
```
Snaipe: I think the note from the original docs is missing, adding a warning like "This should not be used on arrays of struct, consider using cr_assert_arr_eq_cmp instead." could be nice.
(same thing for `cr_assert_arr_neq`)

include/criterion/assert.h
```diff
@@ -30,109 +35,1041 @@
...
+ *
+ *****************************************************************************/
+
 # define cr_expect_arr_neq(...) CR_EXPAND(cr_assert_mem_op_va_(CR_FAIL_CONTINUES_, !=, __VA_ARGS__))
 
-// Safe array comparison assertions
+/**@}*/
+
+/**
+ * @defgroup SafeArrCmpAsserts Safe array comparison assertions
```
Snaipe: There should definitely be something saying that these assertions are C++ and GNU C only. Once in the group, then once per macro would be nice.

include/criterion/assert.h
```diff
@@ -157,19 +1094,144 @@
 
 # ifdef __cplusplus
 
+/**@}*/
+
+/**
+ * @defgroup ExceptionAsserts Exception asserts
```
Snaipe: Same as the safe array assertions, there should be a notice that these are C++-only, even thought it might be obvious.

include/criterion/abort.h
```diff
@@ -42,6 +46,11 @@ CR_API CR_NORETURN void criterion_abort_test(void);
  */
 CR_INLINE static void criterion_continue_test(void) {}
 
+/**
+ *  Kills the current test, marking it as failed.
+ *
+ *  @param[in] msg printf like format string
+ */
```
Snaipe: You're missing an `@param[in] ... format parameters` here.

include/criterion/assert.h
```diff
@@ -30,109 +35,1041 @@
...
+
+/**
+ * Fails always.
+ *
+ * Call: \c cr_assert_fail([FormatString, [Args...]]).\n
+ * The test is marked as failure and the execution of the function is aborted.
+ * The optional string is printed on failure.
+ *
+ * @param[in] fmt printf like format string
+ *
```
Snaipe: All macros should probably have an `@param[in] Args... the format paramaters` or something.

include/criterion/assert.h
```diff
@@ -30,109 +35,1041 @@
...
+ *****************************************************************************/
+
 # define cr_expect_lt(...) CR_EXPAND(cr_assert_op_va_(CR_FAIL_CONTINUES_,  <, __VA_ARGS__))
 
+
+/**
+ * Passes if Actual is less or equal to Reference
+ *
+ * Call: \c cr_assert_leq(Actual, Reference, [FormatString, [Args...]])\n
+ * Passes if Actual is less to Reference.
```
Snaipe: Less or equal to

include/criterion/assert.h
```diff
@@ -30,109 +35,1041 @@
...
+ * @param[in] fmt printf like format string
+ *
+ *****************************************************************************/
 # define cr_assert_leq(...) CR_EXPAND(cr_assert_op_va_(CR_FAIL_ABORT_,     <=, __VA_ARGS__))
+
+/**
+ * Passes if Actual is less or equal to Reference
+ *
+ * Call: \c cr_expect_leq(Actual, Reference, [FormatString, [Args...]])\n
+ * Passes if Actual is less to Reference.
```
Snaipe: Less or equal to

include/criterion/assert.h
```diff
@@ -30,109 +35,1041 @@
...
+ *****************************************************************************/
+
 # define cr_expect_gt(...) CR_EXPAND(cr_assert_op_va_(CR_FAIL_CONTINUES_,  >, __VA_ARGS__))
 
+
+/**
+ * Passes if Actual is greater or equal to Reference
+ *
+ * Call: \c cr_assert_geq(Actual, Reference, [FormatString, [Args...]])\n
+ * Passes if Actual is greater to Reference.
```
Snaipe: Greater or equal to

include/criterion/assert.h
```diff
@@ -30,109 +35,1041 @@
...
+ * @param[in] fmt printf like format string
+ *
+ *****************************************************************************/
 # define cr_assert_geq(...) CR_EXPAND(cr_assert_op_va_(CR_FAIL_ABORT_,     >=, __VA_ARGS__))
+
+/**
+ * Passes if Actual is greater or equal to Reference
+ *
+ * Call: \c cr_expect_geq(Actual, Reference, [FormatString, [Args...]])\n
+ * Passes if Actual is greater to Reference.
```
Snaipe: Greater or equal to

include/criterion/assert.h
```diff
@@ -30,109 +35,1041 @@
...
+/**
+ * @defgroup SafeArrCmpAsserts Safe array comparison assertions
+ * @{
+ */
 
 # if defined(__GNUC__) || defined(__clang__) || defined(__cplusplus)
 
+
+/**
+ * Passes if Actual is comparatively equal to Expected (C++ / GNU C99 only)
```
Snaipe: The `(C++ / GNU C99 only)` should be a full sentence in a note below `The optional string is printed on failure.`, maybe something like:
```
@note
This macro is only available on C++ and GNU C compilers.
```

Snaipe:
I think that the doxygen documentation is a bit too separated from the actual sphinx docs, but this might be better done in a separate issue.

Also, 2 points on the Doxyfile (can't comment inside because there's 2000 lines):

1. exclude `criterion/internal` from the doc generator, otherwise the internals will be picked up by the recursive traversal
2. embedding the README.md might be a bad idea (it renders weird), I would rather have a doxygen-specific main page inside `criterion.h`.

a1lu:
* fixed the issues
* added some notes and warnings
* removed recursive flag in doxyfile
* removed mainpage


Doxygen prints no \n at the end of the comment lines. Would it be nice to have the same format in the rendered file as in the source files?

Snaipe:
As it stands, internal fields in `struct criterion_test_extra_data` are reported, they should be put in a `@internal` section.

> Doxygen prints no \n at the end of the comment lines. Would it be nice to have the same format in the rendered file as in the source files? 

What do you mean? Do you have a specific example in mind?

P.S: You also missed a few issues.

a1lu:
Something like
```
* Otherwise the test is marked as failure and the execution of the function
 * is aborted.
 * The optional string is printed on failure.
```
is printed as 
``
Otherwise the test is marked as failure and the execution of the function is aborted. The optional string is printed on failure.
``
in the html output. So we could add some \n to get the same format. Just a style thing.

Also fixed some Expected -> Unexpected and -> Reference things. 
Your comments above should all be fixed

Snaipe:
Forcing a line feed in those case seems improper -- the next statement should be put in its own paragraph.

That is, instead of:

```
 * Call: \c cr_assert(Condition, [FormatString, [Args...]]).\n
 * Evaluates the condition and passes if it is true.
 * Otherwise the test is marked as failure and the execution of the function
 * is aborted.
 * The optional string is printed on failure.
```

do:

```
 * Call: \c cr_assert(Condition, [FormatString, [Args...]]).
 *
 * Evaluates the condition and passes if it is true.
 * Otherwise the test is marked as failure and the execution of the function
 * is aborted.
 *
 * The optional string is printed on failure.
```

a1lu:
done.

Do you have any notes for the mainpage?

Snaipe:
> Do you have any notes for the mainpage?

Considering that I'll try to merge the doxygen docs with the sphinx docs, there's no need for a mainpage.

Otherwise, everything seems good -- if you don't have anything else to submit, I'll merge this in.

PR #107, opened 2016-02-22
Integrate doxygen with sphinx docs

Snaipe:
Thanks to the recent work of @a1lu, we got some proper doxygen generation on the repository.

However, this effectively split the documentation in half, with sphinx on one hand and doxygen on the other.

This PR aims to harmonize both documentations by integrating doxygen comments into the sphinx doc -- the current result can be browsed [here](http://criterion.readthedocs.org/en/features-sphinx-doxygen).

Notable changes are visible on the assertion reference:

| Before | After |
| --- | --- |
[Rendered](http://criterion.readthedocs.org/en/latest/assert.html) | [Rendered](http://criterion.readthedocs.org/en/features-sphinx-doxygen/assert.html)
[Source](https://raw.githubusercontent.com/Snaipe/Criterion/bleeding/doc/assert.rst) | [Source](https://raw.githubusercontent.com/Snaipe/Criterion/features/sphinx-doxygen/doc/assert.rst)

It is immediately apparent that with these changes, the sphinx documentation isn't a maintainance hell anymore since everything that is API-related can be automatically generated.
This allows us to be thorough with the documentation without sacrificing maintainability (as the sphinx docs will stay in sync with the API docs), which is a huge plus on my side.

As a bonus, the PDF docs also get a pretty nifty [symbol index](https://media.readthedocs.org/pdf/criterion/features-sphinx-doxygen/criterion.pdf#61)

### DIFF
--- a/doc/Doxyfile
+++ b/doc/Doxyfile
@@ -58,7 +58,7 @@ PROJECT_LOGO           =
 # entered, it will be relative to the location where doxygen was started. If
 # left blank the current directory will be used.
 
-OUTPUT_DIRECTORY       = ./doc/doxygen
+OUTPUT_DIRECTORY       =
 
 # If the CREATE_SUBDIRS tag is set to YES then doxygen will create 4096 sub-
 # directories (in 2 levels) under the output directory of each output format and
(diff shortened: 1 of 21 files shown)


am11:
Thank you much for both HTML and PDF versions @Snaipe! :books: 
I have been reading repo docs all this time (which are equally insightful) :+1:

PR #108, opened 2016-02-23
added doxygen for cr_assume_* macros

a1lu:
added doxygen comments for cr_assume macros and grouped them
disabled REPEAT_BRIEF, enabled JAVADOC_AUTOBRIEF @ doxyfile
used the same Argument names as for cr_assert (i.e. Reference, Expected, Unexpected)

### DIFF
--- a/include/criterion/theories.h
+++ b/include/criterion/theories.h
@@ -41,7 +41,10 @@ CR_API void cr_theory_abort(void);
 
 CR_END_C_API
 
-// Theory and datapoint macros
+/**
+ * @defgroup TheoryDatapoints Theory and datapoint macros
+ * @{
+ */
 
 /**
  *  Theory((Params...), Suite, Name, [Options...]) { Function Body }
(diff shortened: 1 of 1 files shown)


doc/Doxyfile
```diff
@@ -107,7 +107,7 @@ BRIEF_MEMBER_DESC      = YES
 # brief descriptions will be completely suppressed.
 # The default value is: YES.
 
-REPEAT_BRIEF           = YES
+REPEAT_BRIEF           = NO
```
Snaipe: These two changes conflicts with the WIP on the sphinx/doxygen harmonization, and doesn't have anything to do with documenting `cr_assume_*` macros.

include/criterion/assert.h
```diff
@@ -149,7 +149,7 @@
 /**@}*/
 
 /**
- * @defgroup CommonBinAsserts Common binary assertions
+ * @defgroup CommonBinaryAsserts Common binary assertions
```
Snaipe: This change conflicts with the WIP on the sphinx/doxygen harmonization, and doesn't have anything to do with documenting `cr_assume_*` macros.

include/criterion/theories.h
```diff
@@ -86,31 +89,201 @@ CR_END_C_API
...
-// Theory invariants
+/**@}*/
+
+/**
+ * @defgroup TheoryInvariants Theory invariants
+ * @{
+ */
 
+/**
+ * Assumes \c Condition is true
```
Snaipe: Use markdown rather than `\c` (i.e. `` `Condition` `` rather than `\c Condition`)

include/criterion/theories.h
```diff
@@ -86,31 +89,201 @@ CR_END_C_API
...
+ *
+ * Call: \c cr_assume(Condition).
```
Snaipe: This line is unecessary for all doxygen comments, as the macro definition matches the API prototype.

include/criterion/theories.h
```diff
@@ -86,31 +89,201 @@ CR_END_C_API
...
+ *
+ * Evaluates \c Condition and passes if it is true.
```
Snaipe: "passes" should really be "continues execution" here.

Ditto for the comments below.

PR #110, opened 2016-03-05
Switched CMAKE_*_DIR variables to PROJECT_*_DIR

MrAnno:
With this change, Criterion can be added as a cmake subproject with the `add_subdirectory()` command.


Snaipe:
Merged manually in `patch` branch, scheduled for 2.2.1.

PR #111, opened 2016-03-06
extended globbing matcher (fix #42)

### LINKED ISSUE #42: Implement an extended globbing parser
Currently, the extended globbing is transformed by the algorithm in `extmatch.c` to a perl regexp, which is then passed to PCRE.

It would probably be better to cut the middle ground and do the parsing ourselves -- this would remove PCRE as a dependency and provide `--pattern` on all platforms.

a1lu:
Implementation of an extended globbing matcher to fix #42 .
The behavior is slightly different to the current implementation:
* Unparenthesized pipes must be escaped too.
* no error_msg string is returned. If a malformed pattern is found, it exits.

Sidenote: cram_test duration decreased from 23.77s to 3.81s


src/string/extglobmatch.c
```diff
@@ -0,0 +1,459 @@
...
+                      break;
+        }
+        p = *con.cur_pat++;
+    }
+    con.ret = 1;
+    return con;
+}
+
+static void malformed_pattern(char p) {
+    printf("pattern error: Unexpected '%c'\n", p);
```
Snaipe: Print the message on stderr.

src/string/extglobmatch.h
```diff
@@ -0,0 +1 @@
+int match(char const *pat, char const *str);
```
Snaipe: `match` is too broad, use `extglob_match` or `extmatch`.

Snaipe:
Remove the old `src/string/extmatch.{c,h}` files. And for that matter, please remove the PCRE dependency in the cmake configuration to truly address #42.

src/string/extglobmatch.c
```diff
@@ -0,0 +1,459 @@
...
+                        if (res == 1) {
+                            return res;
+                        }
+                        eg = find_next(eg);
+                    }
+                    return 0;
+                } else {
+                    malformed_pattern(p);
+                }
+                break;
```
Snaipe: This line is unreachable.

src/string/extglobmatch.c
```diff
@@ -0,0 +1,459 @@
...
```
Snaipe: This line is unreachable.

a1lu:
Removed everything PCRE related I found.

src/string/extglobmatch.c
```diff
@@ -0,0 +1,459 @@
...
```
Snaipe: This line is also unreachable -- either move the return 0 here or replace the break with `/* not reachable */`.

.cmake/Modules/PackageUtils.cmake
```diff
@@ -79,7 +79,7 @@ macro (cr_find_package _PKG)
 
     if (${_PKG_UP}_PKG_FOUND)
       if (${_PKG_UP}_PKG_LIBRARY_DIRS)
-        link_directories(${PCRE_PKG_LIBRARY_DIRS})
```
Snaipe: This seems to be an error in `PackageUtils.cmake`. Replace `link_directories(${PCRE_PKG_LIBRARY_DIRS})` with `link_directories(${${_PKG_UP}_PKG_LIBRARY_DIRS})`.

Snaipe:
Changes in `po/*` are unrelated to this commit -- please do not include them.

a1lu:
* excluded po files
* fixed PackageUtils.cmake
* replaced break with /* not reachable */

a1lu:
* added some tests
* fixed a bug with ? (match any char)

am11:
Hope I don't sound over enthusiastic if I say that I like how the PCRE dependency freedom shapes up :smile:

Thanks for doing this @a1lu! :tada:

src/string/extglobmatch.c
```diff
@@ -0,0 +1,463 @@
...
+                     * A span is either a single character a, or a
+                     * range a-b.  The first span may begin with ']'.
+                     */
+                    do {
+                        unsigned char b = a;
+                        if (a == '\0')	/* Malformed */
+                            goto literal;
+                        if (class[0] == '-' && class[1] != ']') {
+                            b = class[1];
+                            if (b == '\0')
```
Snaipe: I have some trouble understanding this particular check -- and even more so as the branch is never reached.

Since `b` is defined as `a`, and `a` is checked against `\0` @ line 358, this check is in fact redundant and should be removed.

a1lu:
* added test for  src/string/extglobmatch.c:361

test/cram/pattern.t
```diff
@@ -74,17 +98,105 @@ Testing range negation
...
+  [\x1b[0;34m====\x1b[0m] \x1b[0;1mSynthesis: Tested: \x1b[0;34m0\x1b[0;1m | Passing: \x1b[0;32m0\x1b[0;1m | Failing: \x1b[0;31m0\x1b[0;1m | Crashing: \x1b[0;31m0\x1b[0;1m \x1b[0m (esc)
+
+  $ simple.c.bin --pattern 'misc/[!p]assing' --verbose
   [\x1b[0;34m----\x1b[0m] Criterion v2.2.0 (esc)
   [\x1b[0;34m====\x1b[0m] Running \x1b[0;34m2\x1b[0m tests from \x1b[0;33mmisc\x1b[0m: (esc)
   [\x1b[0;33mSKIP\x1b[0m] misc::passing: Test is disabled (esc)
   [\x1b[0;33mSKIP\x1b[0m] misc::failing: Test is disabled (esc)
   [\x1b[0;34m====\x1b[0m] \x1b[0;1mSynthesis: Tested: \x1b[0;34m0\x1b[0;1m | Passing: \x1b[0;32m0\x1b[0;1m | Failing: \x1b[0;31m0\x1b[0;1m | Crashing: \x1b[0;31m0\x1b[0;1m \x1b[0m (esc)
+
+  $ simple.c.bin --pattern 'misc/[' --verbose
```
Snaipe: These should probably go into the "Testing malformed pattern" section of the cram test

src/string/extglobmatch.c
```diff
@@ -0,0 +1,462 @@
...
+    do {
+        if (*pat == '(')
+            paren++;
+        else if (*pat == ')')
+            paren--;
+        ++len; ++pat;
+    } while (paren > 0 && len < patlen);
+    if (paren) {
+        fprintf(stderr, "pattern error: mismatching parenthesis\n");
+        exit(1);
```
Snaipe: 1 as an exit status is used to indicate that the tests failed. It could be wiser to indicate invalid running conditions with a value >= 3.

a1lu:
* fixed nested repeating ext globs
* change exit code to 3
* *() returns -1 for something like  pat: *(a*(b)a) str: aaabbaa ( <-- this should be fixed imo)

src/string/extglobmatch.c
```diff
@@ -0,0 +1,513 @@
...
+    con.ret = 1;
+    return con;
+}
+
+CR_NORETURN static void malformed_pattern(char p) {
+    fprintf(stderr, "pattern error: Unexpected '%c'\n", p);
+    exit(3);
+}
+
+char const *repeat_stack[1000];
```
Snaipe: Private global variables should be marked static. And for that matter, you should avoid global variables altogether and pass the stack pointer to your functions directly (extmatch should be re-entrant).

Snaipe:
If you have specific patterns in mind that cover the remaining 5% coverage, please add them to the tests.

a1lu:
this new version should work like the bash ext globs.

regarding the coverage: 
there is a loop in extglobmatch.c  that is not covered. There was a pattern that covered this loop, but i can't remember this at the moment. If someone find a pattern to cover this (or proof that it is never covered) please contribute.

This implementation is based on [Brzozowski derivative](https://en.wikipedia.org/wiki/Brzozowski_derivative) and some optimised rules [found here, Table 1, section 16.6](https://web.stanford.edu/group/cslipublications/cslipublications/koskenniemi-festschrift/16-watson-frishert-cleophas.pdf)

Snaipe:
Looks nice so far, I'll look into this in detail.

Also, please refactor the symbol names to use snake_case, not camelCase.

src/string/extglobmatch.c
```diff
@@ -0,0 +1,278 @@
...
+    if (pa != paren) {
+        fprintf(stderr,"pattern error: Mismatching parenthesis\n");
+        goto error;
+    }
+
+    catAll();
+    for (; alter > 0; --alter) {
+        glob *b = popFrag();
+        glob *a = popFrag();
+        puts("alter");
```
Snaipe: Remaining `puts`

src/string/extglobmatch.c
```diff
@@ -0,0 +1,278 @@
...
+                  break;
+    }
+    return tmp;
+}
+
+#define pushFrag(f) *frag++ = f
+#define popFrag() *--frag
+
+static glob *compile(char const *pat) {
+    glob *frags[100]={0};
```
Snaipe: the `100` magic number should probably be in a `#define`.

src/string/extglobmatch.c
```diff
@@ -0,0 +1,278 @@
...
+        }
+
+    }
+    if (pa != paren) {
+        fprintf(stderr,"pattern error: Mismatching parenthesis\n");
+        goto error;
+    }
+
+    catAll();
+    for (; alter > 0; --alter) {
```
Snaipe: It looks like this pattern was here to support compound patterns with a mismatching parenthesis, as the only moment where alter would be greater than 0 is because `')'` was not hit. This is already handled L152-L162, so I would drop this loop altogether.

a1lu:
camelCase -> snake_case
removed puts
added define (maybe i should implement a dynamic stack in future)
removed alternative loop

src/string/brz.c
```diff
@@ -0,0 +1,696 @@
...
+        return ret;
+    }
+}
+
+/*
+ * Empty Set
+ * not nullable
+ * no derivative
+ */
+glob *Empty() {
```
Snaipe: Also apply snake_case here. I would recommend adding a prefix like `glob_` to defeat special cases like `Char` or `And` (which is a trigraph on gcc iirc)

a1lu:
renamed Char, And, Or ... -> glob_char, glob_and, ...

test/cram/pattern.t
```diff
@@ -1,20 +1,87 @@
...
+  [\x1b[0;34m====\x1b[0m] Running \x1b[0;34m2\x1b[0m tests from \x1b[0;33mmisc\x1b[0m: (esc)
+  [\x1b[0;33mSKIP\x1b[0m] misc::passing: Test is disabled (esc)
+  [\x1b[0;33mSKIP\x1b[0m] misc::failing: Test is disabled (esc)
+  [\x1b[0;34m====\x1b[0m] \x1b[0;1mSynthesis: Tested: \x1b[0;34m0\x1b[0;1m | Passing: \x1b[0;32m0\x1b[0;1m | Failing: \x1b[0m0\x1b[0;1m | Crashing: \x1b[0m0\x1b[0;1m \x1b[0m (esc)
+
+  $ simple.c.bin --pattern 'misc/[p-f]ailing' --verbose
+  [\x1b[0;34m----\x1b[0m] Criterion v2.2.1 (esc)
+  [\x1b[0;34m====\x1b[0m] Running \x1b[0;34m2\x1b[0m tests from \x1b[0;33mmisc\x1b[0m: (esc)
+  [\x1b[0;33mSKIP\x1b[0m] misc::passing: Test is disabled (esc)
+  [\x1b[0;33mSKIP\x1b[0m] misc::failing: Test is disabled (esc)
```
Snaipe: misc::failing should match here, unless ranges like `[z-a]` arent supported, in which case the test pattern should be `misc/[f-p]ailing` and `misc/[p-f]ailing' should be malformed.

test/cram/pattern.t
```diff
@@ -37,54 +120,119 @@ Selecting both misc::passing and misc::failing
...
+  $ simple.c.bin --pattern '(malformed' --verbose
+  pattern error: Unexpected character '('
+  [3]
 
-  $ simple.c.bin --pattern 'misc/[!azerty]assing' --verbose
+  $ simple.c.bin --pattern 'misc/passing)' --verbose
+  pattern error: Unexpected character ')'
+  [3]
+
+  $ simple.c.bin --pattern 'misc/[' --verbose
```
Snaipe: These tests should be malformed, but appears to be working?

test/cram/pattern.t
a1lu: p-f is not supported and gets rejected by the glob. This is similar to the behavior of my bash

test/cram/pattern.t
a1lu: In Bash something like abc[ matches "abc["
unless you write [..] the [ and ] gets matched es character

test/cram/pattern.t
Snaipe: Then the error message is missing for this case.

test/cram/pattern.t
Snaipe: Fair enough.

test/cram/pattern.t
Snaipe: Oh, I get what you mean -- the pattern gets interpreted literally in bash. I would advise against not putting an error message for this case, as there is no way that a test name would contain brackets.

a1lu: Bash doesn't give an error. It just don't matches.

Snaipe: (Plus it would confuse people for the pattern to silently fail where the error is obviously an human error)

test/cram/pattern.t
Snaipe: I think that mismatching square brackets should also yield an error message, for the same reasons as the `[p-f]` pattern. We don't need to closely follow the spec for these cases, as we work with C identifiers rather than filenames.

test/cram/pattern.t
a1lu: for the sake of simplicity i would support f-p == p-f.

Snaipe: Sure, this seems more sensible.

a1lu:
support for character ranges: c - g == g - c
error for malformed brackets

src/string/extglobmatch.c
```diff
@@ -0,0 +1,276 @@
...
+        case '*': tmp = glob_star(g);
+                  break;
+        case '+': tmp = glob_plus(g);
+                  break;
+        case '@': tmp = glob_at(g);
+                  break;
+        case '?': tmp = glob_opt(g);
+                  break;
+    }
+    return tmp;
```
Snaipe: Reaching this should yield a hard abort(), because there is no way that this can be caused by the user. No need to propagate this error any further.

src/string/extglobmatch.c
```diff
@@ -0,0 +1,276 @@
...
+        int alter;
+        int atoms;
+        char op;
+    } paren[ARRAY_SIZE], *pa;
+
+    int alter = 0;
+    int atoms = 0;
+    pa = paren;
+
+    if (!pat) {
```
Snaipe: Don't bother checking for a `NULL` pattern here, it should be a precondition. Use `assert(pat);`.

src/string/extglobmatch.c
```diff
@@ -0,0 +1,276 @@
...
+                    if (pa == paren || !atoms) {
+                        print_error(p);
+                        goto error;
+                    }
+                    --pa;
+                    cat_all();
+                    a = pop_frag();
+                    tmp = handle_operator(pa->op, a);
+                    free_glob(a);
+                    if(!tmp){
```
Snaipe: This can't reasonably be `NULL`, all pushed operators are among the ones handled by `handle_operator`.

src/string/extglobmatch.c
Snaipe: Uh, let me rephrase that actually: add an `assert(tmp);` because tmp being NULL is not possible.

src/string/extglobmatch.c
```diff
@@ -0,0 +1,276 @@
...
+            break;
+        switch(p) {
+            case '!':
+            case '?':
+            case '@':
+            case '*':
+            case '+':
+                if (*pat != '(')
+                    goto character;
+                if (pa >= paren+100) {
```
Snaipe: This should use ARRAY_SIZE

src/string/brz.c
```diff
@@ -0,0 +1,702 @@
...
+ *
+ * Optimization:
+ * (1) E & Empty -> Empty
+ */
+glob *glob_and(glob const *first, glob const *second) {
+    if (first->type == EMPTY || second->type == EMPTY)
+    {
+        return glob_empty();
+    } else {
+    glob *tmp = new_glob();
```
Snaipe: This block should be indented

Snaipe:
This looks good to me. Once the little indendation nitpick is fixed & the travis build and tests pass, I'll go ahead and merge this.

Thanks for the outstanding work @a1lu!

a1lu:
fixed the indendations, I rebase this if the checks are run.

am11:
👍 Thanks @a1lu!

Should this be removed https://github.com/Snaipe/Criterion/blob/96bf786/.travis.yml#L98-L102? (and we do not need to `brew update` for `brew install llvm cmake` subsequently; saves couple of cycles per run)

Snaipe:
Right, I'll remove that from the config.

PR #114, opened 2016-04-03
Add coverage info to FAQ in docs

blinskey:
First of all, thank you for this terrific project.

I was trying to determine why my tests weren't generating any coverage data, and I found the answer [here](https://github.com/Snaipe/Criterion/pull/84). I didn't see this information anywhere in the official docs, so I added your response to the FAQs in this commit.


Snaipe:
Thank you for this -- I hardly get time to work on criterion this month, so its hard to keep track of issues like this.

PR #117, opened 2016-04-19
XML: The output parameter --xml now generates xml-files with the elap…

kaidowei:
…sed time.

### DIFF
--- a/src/io/xml.c
+++ b/src/io/xml.c
@@ -25,6 +25,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <locale.h>
 #include "criterion/stats.h"
 #include "criterion/options.h"
 #include "criterion/internal/ordered-set.h"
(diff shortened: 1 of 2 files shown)


src/io/xml.c
```diff
@@ -56,7 +57,9 @@
     "failures=\"" CR_SIZE_T_FORMAT "\" "    \
     "errors=\"" CR_SIZE_T_FORMAT "\" "      \
     "disabled=\"" CR_SIZE_T_FORMAT "\" "    \
-    "skipped=\"" CR_SIZE_T_FORMAT "\""
+    "skipped=\"" CR_SIZE_T_FORMAT "\" "     \
+    "time=\"%.3f\""
```
Snaipe: You should use the `g` specifier to not print trailing zeroes

src/io/xml.c
```diff
@@ -112,14 +116,33 @@ const char *get_status_string(struct criterion_test_stats *ts,
     return status;
 }
 
+/*
+ * floats are printed locale dependent, but the xml-specification
+ * requires a dot as the decimal separator.
+ * Therefore we set the locale temporarily to print dots.
+ */
+static int fprintf_locale(FILE *__restrict __stream,
```
Snaipe: use `restrict` instead of `__restrict`, and don't use leading underscore in identifiers

PR #122, opened 2016-05-15
Fix FindCriterion.cmake package name

MrAnno:


Snaipe:
Oh wow. This is embarassing :D 

Thanks for the fix.

PR #124, opened 2016-06-02
Behavior of cr_assert_str* assertions changed:

kaidowei:
- Default assertion messages will now print the expession and the content of both strings.
- String assertions will not crash, if the actual string is NULL.


Snaipe:
@a1lu, could you give a better translation for the new `actual=%s, expected=%s` message part in german?

po/fr.po
```diff
@@ -130,8 +130,10 @@ msgstr "L'expression %s est fausse."
 
 #: src/string/i18n.c:15
 #, c-format
-msgid "The expression (as strings) %s is false."
-msgstr "L'expression (en tant que chaînes de caractères) %s est fausse."
+msgid "The expression (as strings) %s is false: actual=`%s` expected=`%s`."
+msgstr ""
+"L'expression (en tant que chaînes de caractères) %s est fausse: actual=`%s` "
+"expected=`%s`."
```
Snaipe: actual/expected can be translated as "actuel" and "attendu" respectively.

src/string/i18n.c
```diff
@@ -12,11 +12,11 @@ char *cr_translate_assert_msg(int msg_index, ...) {
     static char *messages[] = {
         [CRITERION_ASSERT_MSG_FAIL] = N_("The conditions for this assertion were not met."),
         [CRITERION_ASSERT_MSG_EXPR_FALSE] = N_("The expression %s is false."),
-        [CRITERION_ASSERT_MSG_EXPR_AS_STRINGS_FALSE] = N_("The expression (as strings) %s is false."),
+        [CRITERION_ASSERT_MSG_EXPR_AS_STRINGS_FALSE] = N_("The expression (as strings) %s is false: actual=`%s` expected=`%s`."),
```
Snaipe: I18n strings with multiple printf specifiers must provide positional specifiers. See https://github.com/kaidowei/Criterion/blob/51dcb9a16dab076b748d9b9bb8f2c380a6189caa/src/string/i18n.c#L23-L33 for examples.

po/de.po
```diff
@@ -130,8 +130,9 @@ msgstr "Der Ausdruck %s ist falsch."
 
 #: src/string/i18n.c:15
 #, c-format
-msgid "The expression (as strings) %s is false."
-msgstr "Der Ausdruck (als Zeichenkette) %s ist falsch."
+msgid "The expression (as strings) %s is false: actual=`%s` expected=`%s`."
+msgstr ""
+"Der Ausdruck (als Zeichenkette) %s ist falsch: actual=`%s` expected=`%s`."
```
a1lu: actual / expected in german: aktuell / erwartet

"Der Ausdruck (als Zeichenkette) %s ist falsch: aktuell=`%s` erwartet=`%s`."

src/string/i18n.c
```diff
@@ -12,11 +12,11 @@ char *cr_translate_assert_msg(int msg_index, ...) {
...
         [CRITERION_ASSERT_MSG_IS_NULL] = N_("%s is null."),
         [CRITERION_ASSERT_MSG_IS_NOT_NULL] = N_("%s is not null."),
-        [CRITERION_ASSERT_MSG_IS_EMPTY] = N_("%s is empty."),
-        [CRITERION_ASSERT_MSG_IS_NOT_EMPTY] = N_("%s is not empty."),
+        [CRITERION_ASSERT_MSG_IS_EMPTY] = N_("%s (`%s`) is empty."),
```
Snaipe: The second specifier in this message is useless -- we know it's going to be "(\`\`)" since the message is empty.

po/de.po
Snaipe: Thanks!

kaidowei: @a1lu I'd propose "tatsächlich" instead of "aktuell" because aktuell is used 	in conjunction with time, while tatsächlich is used when talking about reality.

Example: [Was Verbraucher erwarten und was sie tatsächlich bekommen](http://www.heise.de/resale/artikel/Was-Verbraucher-erwarten-und-was-sie-tatsaechlich-bekommen-1845383.html)
roughly translated: What customers expect and what they actually receive

a1lu: @kaidowei AFAIK is the "actual" string the one you want to check against the expected string.
To pass the test the "actual" string (returned from a function, ...) must be match the expected string ( the one you want to be returned by the function).
One could use "tatsächlich" for the expected string, but I think "erwartet" is the better one.

kaidowei: I think, we have a slight misunderstanding.
My translation suggestion is
actual -> tatsächlich
expected -> erwartet

btw: http://sprachen-blog.de/false-friends-aktuell-actual/

a1lu: Yes, I know. 
I thought about the "actual" string as the current string fed into the test.
As you said "aktuell" is used in conjunction with time. So if we take the develop process as a timeline this should be correct.
But I got your point now. I dont know which version is the better one.

Snaipe: "Actual" in this case conveys the meaning of "in reality", more than "current".

a1lu: Ok, then "tatsächlich" is the correct one.


I would not use "tatsächlich" in this case (in spoken language) but something like "Soll / Ist " (target / actual ). Another word could be "vorliegend".

Snaipe:
Looks good to me, I'll re-run the travis tests that didn't start for some reason and if everything is OK, I'll merge this in.

PR #130, opened 2016-07-01
Compilation and test fixes for FreeBSD

kevinoid:
This PR fixes a couple of issues that I encountered when compiling/testing Criterion on a fresh install of FreeBSD 10.3.

The first commit fixes the following build failure:

        Linking C shared library libcriterion.so
        /usr/bin/ld: cannot find -lanl

Which is caused by libanl (part of glibc) not being available.  Since it is only required by nanomsg for `getaddrinfo_a`, I added a check for that symbol to `.cmake/Modules/Capabilities.cmake` and link when present.  This matches the nanomsg behavior.

The second commit fixes a cram test failure in `test/cram/asserts.t` due to the FreeBSD libc defining `NULL` to `nullptr` rathern than `__null` when compiling as C++.  The test now matches against either keyword.

Thanks for considering,
Kevin

### DIFF
--- a/.cmake/Modules/Capabilities.cmake
+++ b/.cmake/Modules/Capabilities.cmake
@@ -26,6 +26,7 @@ check_function_exists(strtok_s HAVE_STRTOK_S)
 check_function_exists(strtok_r HAVE_STRTOK_R)
 
 check_library_exists (rt clock_gettime "time.h" HAVE_CLOCK_GETTIME)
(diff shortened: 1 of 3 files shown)


Snaipe:
Looks good to me. Thank you for your contribution!

kevinoid:
Great.  Thank you for merging it so quickly!

Issue #131, opened 2016-07-06
skip tests


fserra:
Hi

Would it be possible to have something like python's pytest `pytest.skip()` to skip a test at run time?

I am trying to test a software which has different capabilities depending on how it was compiled. So there are functions like `isFunctionalityXAvailable` and would like to write on some tests
```C
if( !isFunctionalityXAvaiable() )
{
    criterion_skip();
}
```
Thanks!

Snaipe:
You can (kind of) already do that by setting up a hook that disables the specific test, but it's too painful to my liking. I guess this will be a workaround for the time being if you *really* need to disable the test, until `criterion_skip` gets implemented.

Thanks for the feedback!

PR #136, opened 2016-07-23
added cr_skip (#131)

a1lu:


include/criterion/internal/assert.h
```diff
@@ -124,7 +125,7 @@ CR_END_C_API
                                                                             \
         CR_STDN free(cr_msg__ - cr_shifted__);                              \
                                                                             \
-        if (!cr_passed__)                                                   \
+        if (!cr_passed__ || Fail == CR_SKIP_ABORT_)                                                   \
```
a1lu: I'm not very happy with this....

Snaipe:
This isn't as easy -- we really want the test to be marked as skipped, not passed.

This implies that for the change to be implemented correctly, you need to:

1. add a SKIP phase kind definition in the [protocol spec](https://github.com/Snaipe/Criterion/blob/bleeding/src/protocol/criterion.proto#L10)
2. regenerate the c support code for the new protocol values by running `make` from within the `src/protocol` directory
3. add a phase handler for this new phase kind in the [client handling code](https://github.com/Snaipe/Criterion/blob/bleeding/src/core/client.c#L378) (see `handle_timeout` for an implementation example)
4. implement cr_skip_test() to send the new SKIP protocol phase message and bail with a call to `exit` (see `criterion_test_die` in `src/core/abort.c` for something similar)

Snaipe:
Also as a side note, there is currently no logger function for skipped tests, you will have to add one by fiddling with `src/log/normal.c`.

a1lu:
To differentiate between the 3 ending states (passed, failed, skipped) I added a enum in  criterion_assert_stats and criterion_test_stats. For the message things an equal enum is used in criterion.proto. So we have the enum two times but I did not want to include the protocol header in the public include files.

Also the skipped test prints a message prints a "[SKIP] Test is skipped" message if skipped, should I hide this? 

Is it necessary to allow an optional user string in cr_skip_test() ?

a1lu:
I merged the old .disabled and the new status SKIPPED so that if a test is skipped on compile time it gets the SKIPPED status like the tests skipped by cr_skip_test().
There is no logger function in src/log/normal.c because the skipped tests are printed by normal_log_post_suite as the tests skipped on compile time.
While in my opinion a on compile time disabled Suite is only a shortcut for disabling all tests by hand i removed the "suite is disabled" string to handle all skipped tests the same

Snaipe:
This seems reasonable so far but do not rename `.disabled` to `.skip` as this breaks backward compatibility. I see no good reason for this change other than aesthetics, and this is really minor in face of all tests written for 2.x and earlier not being able to compile verbatim.

a1lu:
Good point.  I'll undo this.

src/core/stats.c
```diff
@@ -198,14 +193,22 @@ static void push_assert(s_glob_stats *stats,
...
+            ++suite->asserts_failed;
+            ++test->failed_asserts;
+            break;
+        case PASSED:
+            ++stats->asserts_passed;
+            ++suite->asserts_passed;
+            ++test->passed_asserts;
+            break;
+        case SKIPPED:
+            // count skipped tests here?
```
a1lu: Where should we coutn the skipped tests? Here or in line 242?

Snaipe: line 242, as it is a post-exit condition for tests.

src/log/normal.c
```diff
@@ -46,8 +46,7 @@ static msg_t msg_desc = N_("  %s\n");
 static msg_t msg_pre_init = N_("%1$s::%2$s\n");
 static msg_t msg_post_test_timed = N_("%1$s::%2$s: (%3$3.2fs)\n");
 static msg_t msg_post_test = N_("%1$s::%2$s\n");
-static msg_t msg_post_suite_test = N_("%1$s::%2$s: Test is disabled\n");
-static msg_t msg_post_suite_suite = N_("%1$s::%2$s: Suite is disabled\n");
+static msg_t msg_post_test_skip = N_("%1$s::%2$s: Test is skipped\n");
```
Snaipe: For grammatical consistency this should use the past tense.

include/criterion/assert.h
```diff
@@ -50,6 +50,13 @@
  *
  *****************************************************************************/
 # define cr_assert_fail(FormatString, ...) <internal>
+/**
+ * Skips the test
+ *
+ * The test is marked as skipped and the execution of the function is aborted.
+ *
+ *****************************************************************************/
+#define cr_skip_test(void) <internal>
```
Snaipe: Don't specify a macro parameter, this makes `-pedantic` produce a warning when calling `cr_skip_test()`. Just put an empty pair of parenthesis.

Also, use a space between `#` and `define` here.

include/criterion/stats.h
```diff
@@ -30,9 +30,19 @@
 
 # include "types.h"
 
+/*
+ * This enum must be equal to the enum assert.status_type in
+ * src/protocol/criterion.proto
+ */
+enum test_status {
```
Snaipe: This needs to be prefixed

po/fr.po
```diff
@@ -44,44 +44,49 @@ msgid "%1$s::%2$s: Test is disabled\n"
 msgstr "%1$s::%2$s: Le test est désactivé\n"
 
 #: src/log/normal.c:50
+#, fuzzy, c-format
+msgid "%1$s::%2$s: Test is skipped\n"
+msgstr "%1$s::%2$s: Le test est désactivé\n"
+
+#: src/log/normal.c:51
 #, c-format
 msgid "%1$s::%2$s: Suite is disabled\n"
```
Snaipe: This should probably be deleted since the string is gone from the source.

po/fr.po
```diff
@@ -44,44 +44,49 @@ msgid "%1$s::%2$s: Test is disabled\n"
 msgstr "%1$s::%2$s: Le test est désactivé\n"
 
 #: src/log/normal.c:50
+#, fuzzy, c-format
+msgid "%1$s::%2$s: Test is skipped\n"
+msgstr "%1$s::%2$s: Le test est désactivé\n"
```
Snaipe: This should now be `Le test a été sauté\n`.

src/io/json.c
```diff
@@ -102,46 +102,35 @@
...
-
-    const char *status = "PASSED";
-    if (ts->crashed || ts->timed_out)
-        status = "ERRORED";
-    else if (ts->failed)
-        status = "FAILED";
-    else if (is_disabled(ts->test, ss->suite))
-        status = "SKIPPED";
-    return status;
+const char *get_status_string(struct criterion_test_stats *ts){
```
Snaipe: this function should be `static`

Snaipe:
Reading in more detail, I'm not too keen in having the SKIPPED status as a possible result of an assert. Conceptually, you cannot skip a test multiple times (but you can pass or fail multiple assertions, with `expect`).

I would rather have that as a phase transition, just like ABORT and TIMEOUT. In the future 3.0 structure, I could easily imagine having a sub-test being skipped without skipping the complete test.

src/io/json.c
a1lu: It is.

Snaipe: I stand corrected. I think the project really needs something to enforce a coding style.

src/io/tap.c
```diff
@@ -48,15 +48,13 @@ static void print_pre_suite(FILE *f, struct criterion_suite_stats *stats) {
...
-    return t->data->disabled || (s->data && s->data->disabled);
-}
 
 static void print_test_normal(FILE *f, struct criterion_test_stats *stats) {
     const char *format = can_measure_time() ? "%s - %s::%s %s (%3.2fs)\n"
                                             : "%s - %s::%s %s\n";
     fprintf(f, format,
-            stats->failed ? "not ok" : "ok",
+            stats->test_status == FAILED ? "not ok" :
+                stats->test_status == PASSED ? "ok" : "SKIP",
```
Snaipe: this does not follow the [TAP spec](https://testanything.org/tap-version-13-specification.html). For that matter, no need to change anything here -- the SKIPPED case is already handled [L98](https://github.com/Snaipe/Criterion/pull/136/files#diff-0e1bc6cba7767f4d0c3a8eea0dcd5a9aR98)

src/io/xml.c
```diff
@@ -41,7 +41,7 @@
     "tests=\"" CR_SIZE_T_FORMAT "\" "       \
     "failures=\"" CR_SIZE_T_FORMAT "\" "    \
     "errors=\"" CR_SIZE_T_FORMAT "\" "      \
-    "disabled=\"" CR_SIZE_T_FORMAT "\""
+    "skipped=\"" CR_SIZE_T_FORMAT "\""
```
Snaipe: "skipped" is not a valid field for `testsuites` (although it exists for `testsuite`) in the [JUnit 4 XML spec](https://raw.githubusercontent.com/behave/behave/master/etc/junit.xml/junit-4.xsd)

src/io/xml.c
```diff
@@ -56,7 +56,6 @@
     "tests=\"" CR_SIZE_T_FORMAT "\" "       \
     "failures=\"" CR_SIZE_T_FORMAT "\" "    \
     "errors=\"" CR_SIZE_T_FORMAT "\" "      \
-    "disabled=\"" CR_SIZE_T_FORMAT "\" "    \
```
Snaipe: This is part of the spec and must remain.

src/io/xml.c
```diff
@@ -98,22 +97,12 @@
...
-
-    const char *status = "PASSED";
-    if (ts->crashed || ts->timed_out)
-        status = "ERRORED";
-    else if (ts->failed)
-        status = "FAILED";
-    else if (is_disabled(ts->test, ss->suite))
-        status = "SKIPPED";
-    return status;
+const char *get_status_string(struct criterion_test_stats *ts){
```
Snaipe: this should remain `static INLINE`

a1lu: I don't know why github marks "static INLINE" with red, but I added "static INLINE"

include/criterion/internal/deprecation.h
```diff
@@ -32,4 +32,12 @@
 #  define CR_DEPRECATED_(Msg) _Pragma(#Msg)
 # endif
 
+# ifdef __GNUC__
+#   define CR_DEPRECATED_MEMBER(Member) Member __attribute__((deprecated))
+# elif defined (_MSC_VER)
+#   define CR_DEPRECATED_MEMBER(Member) __pragma deprecated (deprecated)
```
Snaipe: This seems to be omitting the member altogether.

a1lu:
This fixup adds optional messages for cr_skip_test()

samples/skip.cc
```diff
@@ -0,0 +1,6 @@
+#include <criterion/criterion.h>
+
+Test(misc, skipping) {
+    cr_skip_test();
+    cr_assert_fail("We want to skip before failing");
+}
```
Snaipe: You forgot to add the `misc::message` test here.

include/criterion/internal/deprecation.h
```diff
@@ -32,4 +32,12 @@
 #  define CR_DEPRECATED_(Msg) _Pragma(#Msg)
 # endif
 
+# ifdef __GNUC__
+#   define CR_DEPRECATED_MEMBER(Member) Member __attribute__((deprecated))
+# elif defined (_MSC_VER)
+#   define CR_DEPRECATED_MEMBER(Member) __pragma deprecated (Member)
```
Snaipe: This still doesn't declare the field on msvc -- you need both the pragma & the field definition. Also, for this case, `CR_DEPRECATED_MEMBER(type foo)` fails as the `deprecated` pragma takes an indentifier, not a definition.

include/criterion/internal/assert.h
```diff
@@ -128,16 +144,33 @@ CR_END_C_API
...
-            (),                                                 \
-            __VA_ARGS__                                         \
+
+# define cr_skip_impl(Skip, ...)                                            \
+    do {                                                                    \
+        char *cr_msg__ = NULL;                                              \
+        int cr_shifted__ = 0;                                               \
+        CR_EXPAND(CR_INIT_SKIP_MSG(cr_msg__, cr_shifted__, __VA_ARGS__));   \
+        Skip(cr_msg__);                                                     \
+        CR_STDN free(cr_msg__ - cr_shifted__);                              \
```
Snaipe: This leaks as `Skip` never returns. This is not solvable as-is, you have to get rid of `cr_skip_impl` entirely as it's overkill for what it does (the only reason this code works for the assert macros is that they also take a default format string and parameters).

I would still implement the optional string as a `"" __VA_ARGS__` as above but make `criterion_skip_test` variadic. The prototype of the function would be:

`CR_API CR_NORETURN void criterion_skip_test(const char *format, ...);`.

Another possibility would be to remove `criterion_skip_test` from the API altogether and make it internal. This way, you have the liberty of freeing the message from the function itself -- but I still think that cr_skip_impl is a bit too complicated for what it really is and that making the function variadic would work best for this purpose.

include/criterion/internal/assert.h
```diff
@@ -104,7 +104,23 @@ CR_END_C_API
         }                                                                      \
     } while (0))
 
+
+
+# define CR_INIT_SKIP_MSG(MsgVar, Shifted, ...) CR_EXPAND(                       \
```
Snaipe: This is now unused -- remove.

include/criterion/internal/assert.h
```diff
@@ -128,16 +144,21 @@ CR_END_C_API
...
+            Fail,                                                           \
+            0,                                                              \
+            dummy,                                                          \
+            CRITERION_ASSERT_MSG_FAIL,                                      \
+            (),                                                             \
+            __VA_ARGS__                                                     \
     ))
 
+# define cr_skip(...)                                                       \
+    CR_SKIP_ABORT_("" __VA_ARGS__);
```
Snaipe: No need to define `CR_SKIP_ABORT_` anymore, just call the function directly.
Also, there's a trailing semicolon here.

src/core/abort.c
```diff
@@ -40,6 +40,28 @@ void criterion_abort_test(void) {
     longjmp(g_pre_test, 1);
 }
 
+void criterion_skip_test(const char *format, ...) {
+    char *msg = NULL;
+    if(strlen(format) > 0)
+    {
+        va_list args;
+        va_start(args, format);
+        cr_asprintf(&msg, format, args);
```
Snaipe: you meant cr_vasprintf.

src/core/abort.c
```diff
@@ -40,6 +40,28 @@ void criterion_abort_test(void) {
     longjmp(g_pre_test, 1);
 }
 
+void criterion_skip_test(const char *format, ...) {
+    char *msg = NULL;
+    if(strlen(format) > 0)
```
Snaipe: just check for `*format`, no need to call strlen

src/core/client.c
```diff
@@ -336,6 +338,19 @@ bool handle_timeout(struct server_ctx *sctx, struct client_ctx *ctx, const crite
     return false;
 }
 
+bool handle_skip(struct server_ctx *sctx, struct client_ctx *ctx, const criterion_protocol_phase *msg) {
+    (void) sctx;
+    (void) msg;
+
+    if (ctx->state < CS_MAX_CLIENT_STATES) {
+        ctx->tstats->test_status = CR_STATUS_SKIPPED;
+        ctx->tstats->message = msg->message ? strdup(msg->message) : 0;
```
Snaipe: use NULL, not 0

test/cram/tap.t
```diff
@@ -96,3 +96,14 @@ Testing CRITERION_OUTPUTS
...
+
+
+  $ skip.c.bin --tap
+  TAP version 13
+  1..2
+  # Criterion v2.2.1
+  
+  # Running 2 tests from misc
+  ok - misc::skipping  # SKIP test was skipped
+  ok - misc::message  # SKIP test was skipped
```
Snaipe: The TAP spec mandates having the skip message after SKIP here. (in other words it should be `Skips may take printf-like messages`)

test/cram/json.t
```diff
@@ -318,3 +318,35 @@ Testing CRITERION_OUTPUTS
...
+            "name": "skipping",
+            "assertions": 0,
+            "status": "SKIPPED",
+            "messages": ["The test was skipped."]
+          },
+          {
+            "name": "message",
+            "assertions": 0,
+            "status": "SKIPPED",
+            "messages": ["The test was skipped."]
```
Snaipe: This is the wrong message

include/criterion/internal/deprecation.h
Snaipe: Further research shows that doing this particular piece of code will, in fact, deprecate the `Member` identifier regardless of the context. Deprecating the word `failed` would be pretty bad, so I think that we'll just have to remove the MSVC-specific part.

include/criterion/internal/assert.h
```diff
@@ -104,6 +104,7 @@ CR_END_C_API
         }                                                                      \
     } while (0))
 
+
```
Snaipe: Don't add unnecessary newlines

Snaipe:
LGTM -- I'll merge this in once the two nitpicks are fixed.

Once again, thanks for your contribution.

Snaipe:
@a1lu I'm going to merge this since it's overall functional, but I'm having an issue with the skip sample when running tests in parallel:

```
$ ./samples/skip.c.bin --verbose
[----] Criterion v2.2.1
[====] Lancement de 2 tests dans misc:
[RUN ] misc::message
[RUN ] misc::skipping
[====] Synthèse: Testés: 0 | Validés: 0 | Échoués: 0 | Plantages: 0 
```

The `[SKIP]` messages are missing.

Could you please fix that in a follow-up PR? I'm merging this now since the "bug" is cosmetic only and fixing conflicts brought by the merging of #135 were a bit of a hassle to fix.

Snaipe:
Rebased and merged as 7143118.

Issue #127, opened 2016-06-21
Debugging a criterion test with GDB


Wiler:
Hi,

I can not find any information about how to debug a criterion test using GDB. When I try to use gdb, it is not capable of show me where the error is produced inside my code. The backtrace only shows me the call path inside criterion. I suppose that GDB needs an special configuration. Could you give me any information related to this?

Thanks

Snaipe:
There are two main ways:

### Single mode

The simplest one is, if you know which test fails, to run your test with the `--single suite::test` command line option. For instance, debugging the `misc::failing` test in the `simple.c` sample:

```gdb
$ gdb -q ./samples/simple.c.bin
Reading symbols from ./samples/simple.c.bin...done.
(gdb) b misc_failing_impl 
Breakpoint 1 at 0x400cc9: file /home/snaipe/workspace/c/criterion/samples/simple.c, line 4.
(gdb) r --single misc::failing
Starting program: /home/snaipe/workspace/c/criterion/build/samples/simple.c.bin --single misc::failing
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".

Breakpoint 1, misc_failing_impl () at /home/snaipe/workspace/c/criterion/samples/simple.c:4
4	    cr_assert(0);
(gdb) q
```

### GDB Inferiors

A more complete/extensive method is to make use of some gdb options to juggle between what gdb calls "inferiors":

```
$ gdb -q ./samples/simple.c.bin
Reading symbols from ./samples/simple.c.bin...done.
(gdb) set follow-fork-mode child
(gdb) set detach-on-fork off
(gdb) set non-stop on
(gdb) set target-async on
(gdb) set schedule-multiple on
```

These options will make sure that the tests are still tracked by gdb without suspending the parent runner. All it left is to place some breakpoints and run the tests:

```
(gdb) b misc_failing_impl 
Breakpoint 1 at 0x400cc9: file /home/snaipe/workspace/c/criterion/samples/simple.c, line 4.
(gdb) r
Starting program: /home/snaipe/workspace/c/criterion/build/samples/simple.c.bin 

<snip>

Thread 2.1 "simple.c.bin" hit Breakpoint 1, misc_failing_impl () at /home/snaipe/workspace/c/criterion/samples/simple.c:4
4	    cr_assert(0);
[Thread 0x7ffff6c0e700 (LWP 7239) exited]
[Inferior 3 (process 7236) exited normally]
```

[...]

Issue #85, opened 2015-12-07
Add an interactive debugging mode


ckaran:
@Snaipe You started talking about adding an interactive mode to Criterion; that sounded sufficiently different from #82 that I decided to make this issue to track it separately.  If you don't think that's a good idea, then please close this issue.

ckaran:
If you're going to make an interactive mode, I'd like to suggest combining this with any work using [nanomsg](http://nanomsg.org/index.html) or any of the other [similar options](http://queues.io/), and then to go ahead and define a full protocol for communications between the parent and children using something like [MsgPack](http://msgpack.org/index.html) as the serialization library.  This would make it possible to turn the children into servers in a client/server architecture, and have any language that is able to work with [nanomsg](http://nanomsg.org/index.html) and [MsgPack](http://msgpack.org/index.html) act as clients.  There a number of advantages to this, including:

- The children/servers can be run on dedicated test hardware while clients/parents can be run on completely separate hardware.  This way if you're doing something that may crash the server, you'll still get logs of what is going on up to when the crash occurred (modulo some lost data due to nanomsg's buffering).

- We can use scripting languages to drive the children and intelligently probe the results.  As an example, I use [Hypothesis](http://hypothesis.readthedocs.org/en/master/) to do what is essentially fuzz testing of my python code.  Using [Cython](http://cython.org/), I've been able to test my C code by making them python extension modules, and then letting [Hypothesis](http://hypothesis.readthedocs.org/en/master/) work its magic.

- By defining a communications protocol, you can separate out the shell/GUI from the tests.  That should keep Criterion a little cleaner (do you *really* want to write an interactive shell for a unit testing framework?)

Snaipe:
This seems fair and would be the general direction to take to allow feature requests like #69 to be implemented.

There are clear issues on the current I/O system that were inherited from a time where it did not really matter, and cleanly rewriting this logic should keep things decoupled between the CLI front end, the server backend, and the test workers.

am11:
[Protobuf](https://github.com/google/protobuf) or [Bond](https://github.com/microsoft/bond) can also be tested for ultra-fast [de]serialization.

Snaipe:
The protocol shouldn't be complicated enough to require something like protobuf -- it would be like swatting a fly with a rocket launcher: highly effective but a tad overkill.

ckaran:
[...]

PR #138, opened 2016-08-06
added gdb and coverage how to

a1lu:
check side note of #127, addition to #114

### DIFF
--- a/doc/debug.rst
+++ b/doc/debug.rst
@@ -1,0 +1,61 @@
+Debugging and Coverage information
+==================================
+
+.. _gdb-ref:
+
+
+Debugging with GDB
+------------------
+
+In one terminal do:
+
+.. code-block:: bash
+
+    $ ./simple.c.bin --debug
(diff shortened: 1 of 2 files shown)


Snaipe:
I'm going to put this on hold until #85 gets fixed. With the upcoming BoxFort integration described in #135, we're going to have a very convenient embedded gdbserver for our tests, so the documentation will need to reflect that accordingly.

Snaipe:
Now that #135 is merged, can you update the debugging section?

The basic usage is:

```
$ ./test --debug
Process test created; pid = 20803
Listening on port 1234
```

```
$ gdb -q ./test
Reading symbols from ./test...done.
(gdb) target remote localhost:1234
Remote debugging using localhost:1234
0x00007ffff7dd9d90 in _start() from target:/lib64/ld-linux-x86-64.so.2
(gdb) q
```

Also see docs for `--debug` and `--debug-transport`.

Snaipe:
Rebased and merged manually as fbbfcb6

Issue #126, opened 2016-06-18
force Criterion to crash on a failed assert


fserra:
For debugging purposes, it would be handy to have some option like `--crash` so that `gdb` gains control when a `cr_assert*` or `cr_expect*` fails

fserra:
this is great,thanks!

PR #139, opened 2016-08-06
added --crash cli option (#126)

a1lu:
Crash the test if an assert fails so that an debugger can gain control.
signal() and abort() should be portable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,6 +55,7 @@ include_directories(
   dependencies/valgrind/include/
   dependencies/klib/
   dependencies/nanopb/
+  dependencies/debugbreak/
 )
 
 # Coverage
--- a/include/criterion/options.h
+++ b/include/criterion/options.h
@@ -116,6 +116,14 @@ struct criterion_options {
      *  default: false
      */
     bool wait_for_clients;
+
+    /**
+     * Raise a debug trap to crash the test if an assert fails so that a
+     * debugger can gain control.
+     *
+     * default: false
+     */
+    bool crash;
 };
 
 CR_BEGIN_C_API
--- a/src/entry/params.c
+++ b/src/entry/params.c
@@ -68,6 +68,8 @@
             "prematurely after the test\n"                  \
     "    --verbose[=level]: sets verbosity to level "       \
             "(1 by default)\n"                              \
+    "    --crash: crash failing assertions rather than "    \
+            "aborting (for debugging purposes)\n"           \
     "    -OP:F or --output=PROVIDER=FILE: write test "      \
             "report to FILE using the specified provider\n"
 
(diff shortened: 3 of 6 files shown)


Snaipe:
I think that `--crash` should do just that -- crash. Using signal shenanigans complicates the whole thing on a portability standpoint as SIGABRT is not even raised by `abort()` and _exit isn't defined in winapi.

Ideally this should just `int $3`, but using compiler builtins like `__builtin_trap` should be prefered.

Maybe introduce a new submodule like https://github.com/scottt/debugbreak and use their `debug_break` macro for this purpose?

a1lu:
The new PR uses https://github.com/scottt/debugbreak as submodule. I had to fork it to fix this error:
With GCC 6.1.1:
debugbreak/debugbreak.h:51:1: error: '__inline__' is not at beginning of
declaration [-Werror=old-style-declaration]

If they merge the PR we can switch the dependencie back to upstream.

CMakeLists.txt
```diff
@@ -46,6 +46,7 @@ cr_add_subproject (nanomsg
 )
 
 cr_add_subproject (wingetopt PATH dependencies/wingetopt CMAKE IF MSVC)
+cr_add_subproject (debugbreak PATH dependencies/debugbreak MAKE)
```
Snaipe: Don't add it as a subproject, it's a header only library that does not need building.

a1lu: fixed this

Snaipe:
Looks good -- I'll wait a couple of days before merging to see if scottt/debugbreak#10 gets merged in the meantime.

src/entry/params.c
```diff
@@ -68,6 +68,8 @@
             "prematurely after the test\n"                  \
     "    --verbose[=level]: sets verbosity to level "       \
             "(1 by default)\n"                              \
+    "    --crash: crashs failing tests so that an debugger" \
+            "can gain control\n"                            \
```
Snaipe: Re-reading this, there are some errors: it should be "crash<e>s failing tests so that a debugger<SPACE>can gain control" (letters in brackets mine).

Maybe this could be reworded as "crash failing assertions rather than aborting (for debugging purposes)".

test/cram/help.t
```diff
@@ -19,6 +19,7 @@ Display the help message
       --always-succeed: always exit with 0
       --no-early-exit: do not exit the test worker prematurely after the test
       --verbose[=level]: sets verbosity to level (1 by default)
+      --crash: crashs failing tests so that an debuggercan gain control
```
Snaipe: Ditto.

include/criterion/options.h
```diff
@@ -116,6 +116,14 @@ struct criterion_options {
      *  default: false
      */
     bool wait_for_clients;
+
+    /**
+     * Raise a signal to crash the test if a assert fails so that an debugger
```
Snaipe: This should be "if an assert" and "that a debugger".

Also, this no longer raise a signal -- this could be reworded as "Raise a debug trap"

a1lu:
https://github.com/scottt/debugbreak/pull/10 got merged and I changed the submodule path to upstream.

PR #151, opened 2016-09-04
Fixed output of skipped tests

a1lu:


src/log/normal.c
```diff
@@ -107,30 +107,31 @@ void normal_log_post_test(struct criterion_test_stats *stats) {
...
-            stats->test->category,
-            stats->test->name,
-            stats->elapsed_time);
+    if(stats->test_status == CR_STATUS_SKIPPED) {
+        if(!stats->message) {
+            criterion_pinfo(CRITERION_PREFIX_SKIP, _(msg_post_test_skip),
+                    stats->test->category,
+                    stats->test->name);
+        } else {
+            criterion_pinfo(CRITERION_PREFIX_SKIP, "%s::%s: %s\n",
```
Snaipe: I just realized but the format string needs internationalization, perhaps use `_(msg_test_generic)` and rename both definitions of `msg_test_abort` to `msg_test_generic`.

test/cram/core.t
```diff
@@ -75,6 +75,7 @@ Testing descriptions
   [\x1b[0;31mFAIL\x1b[0m] misc::failing (esc)
   [\x1b[0;33mSKIP\x1b[0m] misc::skipped: Test was skipped (esc)
   [\x1b[0;34m----\x1b[0m]   This one is skipped (esc)
+  [\x1b[0;34m----\x1b[0m]   Just a failing test (esc)
```
Snaipe: This seems wrong

Snaipe:
The cram tests were fine originally, this change seems to be messing things up with the log output order.

src/log/normal.c
```diff
@@ -105,26 +105,29 @@ void normal_log_post_test(struct criterion_test_stats *stats) {
...
-            stats->test->category,
-            stats->test->name,
-            stats->elapsed_time);
+    if(stats->test_status == CR_STATUS_SKIPPED) {
+        if(!stats->message) {
+            criterion_pinfo(CRITERION_PREFIX_SKIP, _(msg_post_test_skip),
+                    stats->test->category,
+                    stats->test->name);
+        } else {
+            criterion_pinfo(CRITERION_PREFIX_SKIP, msg_test_generic,
```
Snaipe: This should be `_(msg_test_generic)` (the `_` macro calls gettext when i18n is active)

src/log/normal.c
```diff
@@ -105,26 +105,29 @@ void normal_log_post_test(struct criterion_test_stats *stats) {
...
-                        ts->test->category,
-                        ts->test->name);
-            } else {
-                criterion_pinfo(CRITERION_PREFIX_SKIP, "%s::%s: %s\n",
-                        ts->test->category,
-                        ts->test->name,
-                        ts->message);
-            }
-
+        if (ts->test_status == CR_STATUS_SKIPPED) {
```
Snaipe: I think that this chunk has to go, as it still makes things inconsistent in parallel for skipped tests that have a description.

It would be better to get rid of this entirely and call log(pre_test) in skip_disabled().

Snaipe:
Rebased and merged manually as 91250fe

PR #155, opened 2016-09-06
fix: set criterion_options.debug to CR_DBG_NATIVE

a1lu:
criterion_options.debug was not set if no argument was given. According
to the docs the native debugger should used, so this member has to be
set to CR_DBG_NATIVE.


Snaipe:
Nice catch! It seems that the original feature got lost with the small refactor of 0a0574c.

PR #161, opened 2016-09-13
section: Fix "Aborted" due to too many open files

kevinoid:
Running a test binary with many tests and/or assertions, such as the following example:

```C

Test(suite, test) {
    /* Note:  2000 needs to be above current RLIMIT_NOFILE */
    for (unsigned int i = 0; i < 2000; ++i) {
        cr_assert(1);
    }
}
```

on systems with ELF binaries causes the program to abort, usually in `call_report_hooks_*` due to `open_module_self` returning `0` due to `EMFILE` from [`open`](https://github.com/Snaipe/Criterion/blob/c5bfe03/src/compat/section-elf.c#L71).

This PR fixes the issue by calling `close_module` for every successful `open_module_self` call and ensuring that `open_module_self` closes any file descriptors it opens when returning a failure code.

Thanks for considering,
Kevin

### DIFF
--- a/src/compat/section-elf.c
+++ b/src/compat/section-elf.c
@@ -123,7 +123,12 @@ int open_module_self(mod_handle *mod)
         return 0;
 
     mod->fd = fd;
-    return open_module_map(mod);
+    if (!open_module_map(mod)) {
+        close(mod->fd);
+        return 0;
+    }
+
+    return 1;
 }
 
 void close_module(mod_handle *mod)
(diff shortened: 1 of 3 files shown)


Snaipe:
Nice catch! I can't believe I missed that.

kevinoid:
No worries.  It's easy to miss a noop for 2/3 formats which is only an issue for large numbers of asserts/tests.  Ha.

Thanks for the fast review and merge!

PR #167, opened 2016-09-28
Do not create sockets in current directory

stv0g:
In my environment I am not allowed to create a UNIX socket in the current working directory (mounted Docker volume).

I think creating the nanomsg socket in /tmp/ or /var/lib/ is a better practice.


stv0g:
Seems like the two failing Appveyor runs are unrelated to my change.

Issue #165, opened 2016-09-20
Rewrite the assertion API


Snaipe:
The assertion API is a bit clunky as it is. I think that the design of that part of the API needs to change for the better for 3.0, as we've learned quite some things in the past two years.

My biggest gripe with the current design is that it's at the same time too inconsistent and too shallow to be useful for both the user and the runtime (it makes printing actual/expected diffs hard).

I have three proposals for a redesign in mind:
## Proposal 1: use long names and per-type functions:

``` c
Test(assert, full)
{
    /* cr_assert(Cond, [msg, ...]) */
    cr_assert(1);
    cr_assert(1, "Hello %s", "World!");

    /* cr_assert_<cmp>_<type>(A, B, [msg, ...])
     *
     * comparators: eq, ne, lt, le, gt, ge
     * types: i8, i16, i32, i64, u8, u16, u32, u64,
     *        str, wcs, tcs (windows-specific TCHAR)
     *        uchar, schar, uint, sint, ulong, slong,
     *        ullong, sllong, flt, dbl, ldbl, bool
     */
    cr_assert_eq_i8(1, 1);

    /* "cr_assert_<cmp>" will simply use the relevant operator and
     * an appropriate stringification for compilers supporting it. */
    cr_assert_eq(1, 1);

    /* the "mem" type requires an additional size parameter:
     * cr_assert_<cmp>_mem(A, B, Size, [msg, ...])
     */
    cr_assert_eq_mem(&(int){ 1 }, &(int){ 1 }, sizeof (int));

    /* cr_assert_<cmp>_cmp(&A, &B, Cmp, [msg, ...])
     * Must take pointers, and uses a user-provided comparator.
     *
     * Unnecessary on C++ compilers where operator== is used.
     */
    cr_assert_eq_cmp(&(int){ 1 }, &(int){ 1 }, int_cmp);

    /* pointer checks */
    cr_assert_null(NULL);
    cr_assert_notnull(0x1);

    /* C++ exceptions */
    cr_assert_throw({ throw std::exception("test") });
    cr_assert_nothrow({});
    cr_assert_anythrow({ throw std::exception("test") });
}
```

This one is the least risky: we basically take the current API, add some type-specific macros, and make the naming and usage conventions consistent. Nothing too far-fetched, and quite consistent with the C philosophy.
## Proposal 2: cleaner syntax w/ assertion sub-expressions:

``` c
Test(assert, nested)
{
[...]

PR #182, opened 2016-11-27
asserts: Implemented new composable assertion API.

Snaipe:
This PR implements the proposal 2 described in #165. Proposal 3 turned out to be more of a hassle due to how the C language works.

This proposal implements assertions as composable sub-functions
for the `cr_assert` and `cr_expect` macros.

For instance, instead of having `cr_assert_eq(1, 2)`, the user can either
use `cr_assert(eq(1, 2))` or `cr_assert(eq(i32, 1, 2))`. This has the
advantage of simplifying the implementation of the assert macro, but
also brings to the table an improved correctness by letting the user enforce
the type of operands with a type tag, and the possibility to
pretty-print and natively compare user types without c11 or compiler-specific
constructs like _Generic or __builtin_types_compatible_p.

This also lets us handle properly special comparison cases like
ulp-based equality for IEEE floats.


Snaipe:
Quick update: I'm still working on those changes, the main challenge here is to integrate a proper diffing algorithm to the codebase for comparing files, large strings, and hex dumps. So far there are very few libraries that does that while being portable (on all supported criterion platforms), documented, and license-compatible.

I may have to implement that myself.

am11:
Not sure if it is viable to take dependency on one of these, but might worth checking out their implementation:
https://github.com/git/git/tree/master/xdiff and
https://github.com/libgit2/libgit2/tree/master/src/xdiff

Snaipe:
Seems like the xdiff implementation of libgit2 is lgpl. I think it could work to create a separate repository just for the xdiff stuff, and link it as a static lib in criterion.

am11:
Just tested it on windOS 😸 

```cmd
git clone https://github.com/libgit2/libgit2 \temp\libgit2
mkdir \temp\build
cd \temp
:: create CMakeLists.txt outside of libgit2 (one level up)
notepad CMakeLists.txt
```

and paste (shaved off all unrelated stuff and we end up with this):

```cmake
PROJECT(libgit2diff C)

# Include POSIX regex when it is required
IF(WIN32 OR AMIGA OR CMAKE_SYSTEM_NAME MATCHES "(Solaris|SunOS)")
	INCLUDE_DIRECTORIES(libgit2/deps/regex)
	SET(SRC_REGEX libgit2/deps/regex/regex.c)
ENDIF()

INCLUDE_DIRECTORIES(libgit2/src libgit2/include)
FILE(GLOB SRC_GIT2 libgit2/src/xdiff/*.c)
ADD_LIBRARY(git2diff ${SRC_GIT2})
```

Finally `cd \temp\build && cmake .. && cmake --build .` emits:

```cmd
C:\temp\build>ls Debug
git2diff.lib  git2diff.pdb
```

am11:
@Snaipe, going by the license doc: https://www.gnu.org/licenses/lgpl.html, if the Criterion repo takes direct submodule dependency on libgit2 and the altered CMakeLists.txt for libgit2 is kept in Criterion repo that only builds xdiff and statically link to libCriterion (LGPL doc refers this as "part of" Library); would it still be abiding by the license terms?

Is it necessary to create a separate repo (like this one: https://github.com/spotrh/libxdiff) for MIT-LGPL bridging? Perhaps, an explicit third-party notice can be added quoting the terms of libgit2 license to meet the requirements.

Snaipe:
It might be a bit more complicated than that, as we can't static link against a LGPL library (or at least we have to provide users a way to use their own version of libxdiff through dynamic linking). libgit2 seems to be under the GPL2 with a linking exception that allows both dynamic & static linking, but they include the xdiff sources directly, which are LGPL. I'm not sure how libgit2 legally handles it.

The best course of action here would be to separate xdiff into its own repository, dynamic link to it, and ship it along with libcriterion.so. I'm not sure that using the (old) xdiff that ships with distributions is going to work, so we might have to name it something like libxdiff-git2.so.

Snaipe:
Now that I think about it, statically linking against our version of xdiff should comply with the LGPL, as long as we provide ways to switch libraries.

I think that introducing a cmake-level option like `-DXDIFF_PATH=/path/to/lib` would be the best course of action. We could also handle the special `pkgconfig` value to signify that pkg-config should be called to get the cflags & ldflags, respectively.

Snaipe:
It turns out that I was able to use the stock libgit2 in criterion -- I initially missed a function that allowed me to diff arbitrary buffers without using a repository.

The new assertions are shaping up nicely!

criterion_new_asserts

am11:
This looks really fantastic! 👏 

Just wondering, if say we are asserting on large file content (say huge JS code files for example), would it show the diff's hunks of single assert individually? If it does (or even doesn't), then would it make sense to have some limit option from get go; say `--diff-limit` with possible values: `hunk<m>..hunk<n>`, `length<n>`, `lines<m>..line<n>`  such that `--diff-limit hunk3` will show all hunks from 3 to last, `--diff-limit hunk8..hunk10` will show hunk 8 to 10, `--diff-limit length220` will truncate everything after 220 characters and similarly by line `--diff-limit line3..line21`.

Reason for this more granular limit option proposal is some CI services impose limit on log output. For instance, I have seen TravisCI jobs choking with:

> The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).
>
> The job has been terminated

(redirecting `std{out,err}` of the command output to a file/device in such scenario is an option but it is a meh workaround tbh)

Snaipe:
What criterion currently does is, for an `eq` operation, that it first transforms the `expected` and the `actual` parameters into strings, makes a diff out of those generated strings, and sends it back to the runner, which ultimately prints it -- so yes, the complete diff is printed for a failed assertion.

As the diff is sent as a single string to print known as the "formatted string" in the protocol, which may contain a formatted representation of the failure, it might be easier & more consistent to simply limit the number of lines taken by this formatted representation, per assertion. Limiting by characters might produce in this case a less coherent output, so I would be more in favor or wrapping the lines to say 80 columns (or the terminal width) and let the line limit do its job.

Limiting by hunks is also a good idea, though I'm not seeing the utility of something like `hunk8..hunk10`. I would see something like `--limit-diff-hunks=<n>` where `<n>` is a signed integer, where positive values show the first `n` hunks, and negative values the last.

Though I wonder if micro-managing the limits of each of these is really a good idea alone. I would see it as an addition to something perhaps more simple, like verbosity levels. I think that revamping the logging system to have a more consistent view of everything by switching verbosity levels could be nice: we could set by default to only show the first hunk of the diff and limit the formatted string to 10 lines, and increase those limits the higher the verbosity level is (+ eventually not put any limits for the highest verbosity level).

Snaipe:
I feel that this PR is almost ready. Pretty much all tags have been implemented, and it's starting to look like what I had in mind. It's by no mean perfect, but since this is opt-in by including `criterion/new/assert.h`, it leaves us some time to fix bugs and make the system more solid as times goes on.

What remains now is:

* [ ] Update the documentation. `criterion/new/assert.h` should have some doxygen, and the assertion reference needs to be updated.
* [ ] Add some helper functions to transform files/streams to a `struct cr_mem` and/or a `char *`, to replace the assertions functions in `criterion/redirect.h` (the plan is to use the `mem`, `str`, and `wcs` tags to do the comparisons, which conveys intent better)
* [ ] Update the json/xml/tap reporters with the new messages
* [ ] Fix the cram tests since the output changed

Snaipe:
Rebased and merged in d258a8b.

Snaipe:
I decided to merge this a little bit early to start iterating on it. The last two points of my bullet list can be addressed later, and so far these changes pass all tests, so it should be good to go.

PR #188, opened 2017-02-17
Master

XyjVam:
The last two commits made it possible for me to use the library.


Snaipe:
LGTM, but we don't accept pull requests on master, only bleeding.

I'll manually cherry pick those commits once I get back home.

Snaipe:
Manually cherry-picked on bleeding as fcaa0d6 and 0891f79.

PR #195, opened 2017-03-18
doc: Fix minor errors in debug.rst.

skytreader:
'debugging server installed on your server' -> 'debugging server installed on
your machine'. Considered changing the first mention of "server" to "software"
but noticed that the "software" is more often called "server" (gdb, lldb). So I
changed the second "server" instead as a matter of style (also, Criterion won't
always be used in a _server_).

'remote target localhost:1234' -> 'target remote localhost:1234'. The given
command is just wrong and unrecognized by gdb.

### DIFF
--- a/doc/debug.rst
+++ b/doc/debug.rst
@@ -6,7 +6,7 @@ Debugging and Coverage information
 .. note::
 
     The following section assumes you have the relevant debugging server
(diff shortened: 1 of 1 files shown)


Snaipe:
LGTM. Thank you for your contribution!

PR #199, opened 2017-04-02
Update CMakeLists.txt

gabrielvv:
Hi, I think there is a minor error in this file.
When running ctest it will never pass because the name of the executable specified with the "add_test" command (2nd argument) is wrong.


Snaipe:
Right -- thanks for catching this!

PR #231, opened 2017-11-21
Make cr_file_match_str fail if str is longer.

Ethiraric:
When using `cr_file_match_str`, if the provided string we want to match the file against is longer than the contents of the file, the function returns that the file matches the string.

This function is used by many other assertion routines (namely those asserting the contents of the standard output streams). This can be very error-prone when designing tests such as the following:

```c
Test(criterion, cr_file_match_str, .init = cr_redirect_stdout)
{
  printf("%s%d%s", "string", 42, "other");
  fflush(stdout);
  cr_assert_stdout_eq_str("string42othernonsense");
}
```

This commit makes it so that `cr_file_match_str` detects this kind of mismatch.

Note: It does not alter the fact that the function consumes all of the stream.

### DIFF
--- a/samples/redirect.cc
+++ b/samples/redirect.cc
@@ -32,7 +32,7 @@ void rot13_io(void)
 {
     std::string s;
 
-    std::cin >> s;
+    std::getline(std::cin, s);
     for (size_t i = 0; i < s.length(); ++i)
         s[i] = rot13_char(s[i]);
     std::cout << s << std::flush;
--- a/src/io/file.c
+++ b/src/io/file.c
@@ -22,6 +22,11 @@ int cr_file_match_str(FILE *f, const char *str)
     /* consume the rest of what's available */
     while (fread(buf, 1, sizeof (buf), f) > 0) ;
 
+    /* there are more bytes in str than in f */
+    if (len) {
+        return 0;
+    }
+
     return matches;
 }
 


Snaipe:
Thanks for the contribution. The idea is sound.

I'll merge that once the code and/or the tests are fixed so the CI builds pass.

src/io/file.c
```diff
@@ -22,6 +22,10 @@ int cr_file_match_str(FILE *f, const char *str)
     /* consume the rest of what's available */
     while (fread(buf, 1, sizeof (buf), f) > 0) ;
 
+    /* there are more bytes in str than in f */
+    if (len)
```
Snaipe: Stylistically, this should be:

```
if (len) {
    return 0;
}
```

Ethiraric:
The issue with the CI seems to be unrelated to my modifications.

```git checkout -qf %APPVEYOR_REPO_COMMIT%
fatal: reference is not a tree: 2ac40c5b661a7f02d85bef545c4036b126c70659
Command exited with code 128```

I'll have a look at the Travis output, which seems to be more verbose.

File: src/io/file.c (hunk 1 above)
Comment (Ethiraric) [reply src/io/file.c]: I'll patch that right away.

**Comment by Ethiraric** [comment]
My understanding of the unit tests seems to not be on point.

Do I just have to `cmake ${ROOTDIR} -DCTESTS=ON` and then build and run `ctest`?

Everything but the "cram" test works fine on my machine (which was already failing the commit before).

Is there any extra configuration I need to set?

**Comment by Snaipe** [comment]
You need cram for running the system tests. You can install it with `pip`/`pip3` depending on your python installation.

**Comment by Ethiraric** [comment]
Everything should be alright now. I've installed `cram` and all tests succeed now.

One of the tests had to be changed. I've updated the commit message to reflect that change.

File: samples/redirect.cc [hunk 2]
```diff
@@ -32,7 +32,7 @@ void rot13_io(void)
 {
     std::string s;
 
-    std::cin >> s;
+    std::getline(std::cin, s);
```
Snaipe: That's a great catch! Since we didn't inspect the rest of the string before, this went unnoticed.

Snaipe:
Don't mind the Travis stylistic fail -- that's just the uncrustify config being broken. We're switching to clang-format, but in the meantime I'm reviewing style manually.

I don't know what's up with Appveyor -- it worked recently. I'll take a look at it.

PR #237, opened 2018-01-30
Look for libintl headers in the correct location

opoplawski:
If libintl.h is not in the default search path, we need to look for it in the right location.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -149,6 +149,7 @@ endif()
 # I18N
 
 if (I18N AND GETTEXT_FOUND AND LIBINTL_LIB_FOUND)
+  include_directories(SYSTEM ${LIBINTL_INCLUDE_DIR})
   include(GettextTranslate)
   add_subdirectory(po)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -149,6 +149,7 @@ endif()
 # I18N
 
 if (I18N AND GETTEXT_FOUND AND LIBINTL_LIB_FOUND)
+  include_directories(${LIBINTL_INCLUDE_DIR})
```
Snaipe: Please make that a system include with `include_directories(SYSTEM ...)`. It's not likely that the libintl headers are going to introduce warnings on their own, but it's a bit more correct.

Snaipe:
Could you rebase the branch on top of bleeding, so a new build can be re-triggered? (There were some CI fixes that were needed)

opoplawski:
Rebased and added SYSTEM

Snaipe:
Rebased and merged as 45f6f41.

PR #252, opened 2018-05-14
Fixes stream.h include so it can be found.

mlopezqc:
While working on another project I started having issues with your latest code. I found that the recently added stream.h was not added to the CMakeList.txt and it was not well referenced in the assert.c file. I fixes those for my project.

### DIFF
--- a/include/criterion/new/assert.h
+++ b/include/criterion/new/assert.h
@@ -433,7 +433,7 @@
 #endif /* !CRITERION_DOCGEN */
 
 #include "memory.h"
-#include "../internal/assert/stream.h"
+#include "stream.h"
 
 /* Internals */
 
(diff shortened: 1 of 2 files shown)


Snaipe:
Good catch for the CMakeLists.txt bit, but I don't get why the `#include` needs to be changed; the header is in `criterion/new`, so it should be including `criterion/new/stream.h`. Is it not working on your end?

Snaipe:
All the builds are failing with that change, so I don't think it's the right fix. What was the original problem that prompted this fix?

stdevMauricio:
I seems to me that not including stream.h in the CMakeList was the source of the errors. There are two of them, so I got confused. I'd recommend you to delete the src/include folder before running the tests, so the compiler is forced to use the installed libraries and early detect if there's something missing in CMakeList. Sorry if I didn't make an issue, but I'm really short on time. Just hope this helps.

Snaipe:
Alright, looks good to me (the new build errors are unrelated). Thanks for the fix!

Snaipe:
Rebased & merged as 97d6ec7.

PR #254, opened 2018-05-29
Add missing parameter to the documented interface.

siebenschlaefer:
cr_(assert/expect)_mem_(n)eq are implemented by calls to
cr_assert_mem_op_va_. But while the documentation of
cr_(assert/expect)_mem_neq contains the third parameter Size, the
documentation for cr_(assert/expect)_mem_eq does not.

This fix only touches the "declaration" and the documentation of
cr_assert_arr_eq and cr_expect_arr_eq.

### DIFF
--- a/include/criterion/assert.h
+++ b/include/criterion/assert.h
@@ -1101,11 +1101,12 @@
  *
  * @param[in] Actual Array to test
  * @param[in] Expected Expected array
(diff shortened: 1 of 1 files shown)


Snaipe:
Good catch! Thanks for the fix.

PR #259, opened 2018-06-23
Add directory support to test reports

MrAnno:
This PR adds directory support to test reports (`CRITERION_OUTPUTS`, `--output`).

The functionality is the same as in GTest (GTEST_OUTPUT):

`--output=PROVIDER:PATH`: Write a test report to `PATH` using the output provider named by `PROVIDER`.
If `PATH` is an existing directory, the report will be created in that directory.

The report file will be named after the binary. For example: test_foo --> test_foo.xml

If the file already exists, Criterion will pick a different name (test_foo_1.xml, test_foo_2.xml) to avoid overwriting it.

----------------------

Adding a dynamically sized string class to Criterion was an enormous overkill, but it made the implementation of `cri_path_gen_unique_filename()` and `cri_path_combine()` much cleaner.

I hope I haven't broken Criterion's KISS principle too much. :(

Questions:
- Currently, the generated file extension is the name of the provider. Is that OK? Should I add a `default_extension`) parameter to `criterion_register_output_provider`?

Improvement ideas:
- The parameter list of `cri_path_build()` is NULL-terminated.
  I'm thinking of creating a macro `#define cri_path_combine(...) cri_path_build(CRI_PATH_SEPARATOR, ##__VA_ARGS__, NULL)`, but this is not standard C99.
  I've read your [blog post](https://snai.pe/c/preprocessor/varargs/) about varargs. I think `Length parameter + Macro` is what I need here.
  Is there an existing implementation of this approach in Criterion that I can use?

--------------------

- [x]  The new string facility seems a bit redundant here; I think you might be better off using cri_fmt_bprintf(&buf, &offset, &size, <format>, ...), which essentially appends to buf the formatted string, resizing it as needed.

- [x]  Some of those commits don't really make sense by themselves. I think it would be best to squash those (except the appveyor one, because this one is pretty much standalone)

- [x]  Remove the stylecheck commit -- I'll address that in another change, see below.

### DIFF
--- a/include/criterion/internal/asprintf-compat.h
+++ b/include/criterion/internal/asprintf-compat.h
@@ -43,6 +43,8 @@ CR_API void cr_asprintf_free(char *buf);
 
 CR_API int cri_fmt_bprintf(char **buf, size_t *offset, size_t *sz,
         const char *fmt, ...);
+CR_API int cri_fmt_vbprintf(char **buf, size_t *offset, size_t *sz,
+        const char *fmt, va_list ap);
 
 CR_API char *cri_strtok_r(char *str, const char *delim, char **saveptr);
 
--- a/include/criterion/options.h
+++ b/include/criterion/options.h
@@ -218,6 +218,8 @@ struct criterion_options {
      * default: false
      */
     bool ignore_warnings;
+
+    const char *executable_name;
 };
 
 CR_BEGIN_C_API
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -33,8 +33,8 @@ set(SOURCE_FILES
   src/compat/strtok.h
   src/compat/process.c
   src/compat/process.h
-  src/compat/basename.c
-  src/compat/basename.h
+  src/compat/path.c
+  src/compat/path.h
   src/compat/mockfile.c
   src/compat/time.c
   src/compat/time.h
--- a/src/compat/path.c
+++ b/src/compat/path.c
@@ -23,8 +23,10 @@
  */
 #include <stdlib.h>
 #include <string.h>
+#include <stdarg.h>
 
 #include "path.h"
+#include "err.h"
 #include "config.h"
 #include "string/fmt.h"
 
@@ -33,6 +35,21 @@
 # include <limits.h>
 #elif defined (HAVE_GETCURRENTDIRECTORY)
 # include <windows.h>
+#endif
+
+#if defined (_WIN32) && !defined (__CYGWIN__)
+# include <windows.h>
+#else
+# include <unistd.h>
+
+# include <sys/types.h>
+# include <sys/stat.h>
+#endif
+
+#if defined (_WIN32) || defined (__CYGWIN__)
+# define CRI_PATH_SEPARATOR '\\'
+#else
+# define CRI_PATH_SEPARATOR '/'
 #endif
 
 const char *basename_compat(const char *str)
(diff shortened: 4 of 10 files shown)


Snaipe:
Thanks for the contribution! The patch is a bit large, so it might take a bit for me to review it completely.

Some general comments:

* The new string facility seems a bit redundant here; I think you might be better off using `cri_fmt_bprintf(&buf, &offset, &size, <format>, ...)`, which essentially appends to `buf` the formatted string, resizing it as needed. I think it's a good time for me to document a few of those internals.
* Some of those commits don't really make sense by themselves. I think it would be best to squash those (except the appveyor one, because this one is pretty much standalone) 
* Remove the stylecheck commit -- I'll address that in another change, see below.

> Currently, the generated file extension is the name of the provider. Is that OK? Should I add a default_extension) parameter to criterion_register_output_provider?

Yes, that makes sense.

> The parameter list of cri_path_build() is NULL-terminated.
I'm thinking of creating a macro `#define cri_path_combine(...) cri_path_build(CRI_PATH_SEPARATOR, ##__VA_ARGS__, NULL)`, but this is not standard C99.
I've read your blog post about varargs. I think Length parameter + Macro is what I need here.
Is there an existing implementation of this approach in Criterion that I can use?

I think there's some shady stuff doing that in `criterion/internal`, but I wouldn't poke at that hornet nest to be completely honest. I'd rather have a macro that doesn't stringify __VA_ARGS__, and we'd just pass strings instead.

> The PR contains a commit that allows stylecheck to fail in Travis CI. The bleeding branch has ~1000 lines of code that are not accepted by uncrustify.

Yes. Uncrustify has been a buggy mess ever since I introduced it, I'm currently considering switching to something else (maybe  `clang-format`?). Just ignore the style check job, I only check whether the actual builds pass anyway.

> A test (alloc.cc) is not compiled at all, but I couldn't enable it due to C++-only problems in the new assertion API (internal/assert/op.hxx)

Could you elaborate on the problems you encountered? I can't believe I forgot to add `alloc.cc` in the tests, Geez.

Snaipe:
> The new string facility seems a bit redundant here; I think you might be better off using cri_fmt_bprintf(&buf, &offset, &size, <format>, ...), which essentially appends to buf the formatted string, resizing it as needed. I think it's a good time for me to document a few of those internals.

Looking back, `cri_fmt_bprintf`, which somewhat meant `buffer printf`, is really not aptly named. I think `cri_fmt_appendf` would have made more sense, or something of that effect.

MrAnno:
I'm going to reduce the size of the PR and address the review notes. Thank you for the initial review :)

MrAnno:
I've just replaced every `cri_string` operation with `cri_fmt_bprintf()`.

The end result looks like this (compared to the original version):
https://gist.github.com/MrAnno/9aaa0ded3922bacacbeb11f7d6a85e8d

I think `cri_path_build_va()` became more complex because of the noise that the `output, offset, size` triplet generates. I also had to use the `"%.*s"` format string to imitate the behavior of `cri_string_append_length()`.

Please have a look at it. If you prefer the new version, I'll update the pull request.

Snaipe:
Perhaps the three first parameters are redundant, but I'd rather have the changes somewhat minimal and atomic. I agree that having to specify those might be a bit tiring, so perhaps it would be indeed best to introduce a slice-like type:

```
struct cri_slice {
    void *data;
    size_t len;
    size_t cap;
};
```

and change bprintf so that it resizes the slice if the capacity isn't enough, and appends at `(char*)data + len`. At least, I think that such a slice type would be more useful for most applications than just a string type.

MrAnno:
Sounds good, I've updated the PR.

PR #286, opened 2019-02-02
print name of uncaught signals

a1lu:
uses built-in kill command of sh to generate the list of signal names. If
sh is not available the signal number is printed

refs #128


src/signals/create_source.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
```
am11: Maybe we can simply implement it in C by simply using `#include <signal.h>`, without shelling out the introspection. e.g.:
https://github.com/karelzak/util-linux/blob/917f53cf13c36d32c175f80f2074576595830573/lib/signames.c#L55-L132

a1lu: We could do that if we only support posix signals.

a1lu: @Snaipe Whats your opinion on this? Posix or the local signals?

PR #288, opened 2019-02-16
Struct pointer references use -> instead of dot

szaydel:
This seems like a documentation bug. I cannot see how this can be correct in its current form.


Snaipe:
LGTM. Thanks for the change!

PR #296, opened 2019-06-18
Adding install instructions for fedora and centos

samber:


Snaipe:
I'm not against users providing RPMs for Criterion, but I think we need to mention that this is 1) a user package and 2) it'd be better if it was platform-agnostic.

An even better thing to do would be to reach out fedora package maintainers and make a request to have criterion included in their repositories. Upstream maintainers usually can't do that themselves as far as I can tell, so if anyone would like to see Criterion included on fedora & derivatives, please reach out to them. :)

samber:
Yes, it was my plan initially. But I did not have time to go further.

RPM fusion is probably the best repo to publish such a package.
What do you think ?

Snaipe:
That works for me.

Are there any gotchas for building Criterion on fedora-based distros? If there are, I'll also think about adding a build job to the CI based on a fedora docker image.

MrAnno:
Closing due to inactivity.

Criterion is still not packaged on Fedora/CentOS/Rocky/Alma, it might be worth giving it a shot on Fedora.
This can be tracked in #321.

PR #300, opened 2019-09-03
WIP: Handle exit status changes after the end of a test

### LINKED ISSUE #220: The runner ignores exit statuses after the test is done executing
After the discussion with @cbatten on #204 about address sanitizer, I went on to test my allegations since it's been a few releases since I haven't checked that particular synergy.

It turns out that the introduction of the new client/server code broke it. In particular, [this bit](https://github.com/Snaipe/Criterion/blob/bleeding/src/core/client.c#L311-L321) reports the test success (or failure), but this happens while the worker is still alive -- and while an `exit` called in the test is going to trigger the death event, and fail the test, Address Sanitizer and Valgrind on the other hand exit with a non-zero exit status after the main has done executing, which is well after the success or failure of the test has been reported and printed.

The incorrect condition came from the change of semantics in the runner's protocol: before the refactor, the worker process *was* the test itself. Now, the runner conceptually supports workers that may run more than one test.

I think the best fix for that is to simply restrict the procotol to only accept one test per process, and then handle the reporting after the worker exits. Additional tests can always be run as sub-tests of the process "root" test.

Fumesover:
I'm sure there is a better way to fix that, but it works and allow you to detect if the test crashed after the end of a test.
This fix creates an edge effect, if an exit test in the init or finish, it will be marked as FAILED instead of PASSED.
This can resolve #220 

Edit: Need more work to integrate this

### DIFF
--- a/src/core/client.c
+++ b/src/core/client.c
@@ -494,6 +494,15 @@ bool handle_death(struct server_ctx *sctx, struct client_ctx *ctx, const criteri
                 break;
             }
             ctx->tstats->exit_code = death->status;
+            if (ctx->tstats->test->data->exit_code != death->status &&
+                    ctx->tstats->test_status == CR_STATUS_PASSED) {
+                --ctx->gstats->tests_passed;
+                --ctx->sstats->tests_passed;
+                ++ctx->gstats->tests_failed;
+                ++ctx->sstats->tests_failed;
+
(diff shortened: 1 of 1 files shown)


Fumesover:
Wrong approach, it can't work like that. (Logs cannot work with this patch)

PR #310, opened 2019-10-31
Update the share directory location

kdridi:
This location depends on the initial install directory of automake. Eg, it can be installed inside /usr/local instead of /usr

### DIFF
--- a/dev/autotools/autogen.sh
+++ b/dev/autotools/autogen.sh
@@ -2,9 +2,10 @@
 
 # copy TAP driver into build-aux
 automake_ver=$(automake --version | \grep -E -o '[0-9]\.[0-9]{2}')
+automake_dir=$(dirname $(which automake))/..
 
 mkdir -p build-aux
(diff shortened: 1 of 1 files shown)


Snaipe:
LGTM. Thanks for the fix!

dev/autotools/autogen.sh
```diff
@@ -2,9 +2,10 @@
 
 # copy TAP driver into build-aux
 automake_ver=$(automake --version | \grep -E -o '[0-9]\.[0-9]{2}')
+automake_dir=$(dirname $(which automake))/..
```
am11: Maybe `command -v automake` is more portable.

Snaipe:
I might have prematurely merged this since I thought it was pretty trivial, but even beyond `command -v` and `which` shenanigans, I've since realized that the obvious solution here would have been to simply use `automake --print-libdir`. :man_facepalming: 

I'll fix that this weekend unless someone beats me to it with a PR.

PR #318, opened 2019-12-28
theories: switch from dyncall to libffi

Snaipe:
As much as I like dyncall, it just isn't that broadly available on linux
distros. Using libffi has the advantage of being ubiquitous and targets
all of our platforms.

So long, and thanks for all the fish.

---

Some questions remain before this can be merged. Notably, I need to figure out the best way to vendor libffi on a naked windows install (as in, without msys2 or other package managers). A cursory glance seem to indicate that libffi uses autotools. I might need to write a meson wrap file.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -129,11 +129,9 @@ threads = dependency('threads')
 nanomsg = dependency('nanomsg')
 libgit2 = dependency('libgit2')
 boxfort = dependency('boxfort', fallback: ['boxfort', 'boxfort_dep'])
+libffi = dependency('libffi', required: get_option('theories'))
 
-# These libraries are currently broken for not providing a pkg-config file
-dyncall = cc.find_library('dyncall_s', required: get_option('theories'))
-
-config.set('ENABLE_THEORIES', dyncall.found())
+config.set('ENABLE_THEORIES', libffi.found())
 
 # optional platform-dependent standard libs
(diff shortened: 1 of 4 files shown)


am11:
> write a meson wrap file.

seems like someone has done the work: https://github.com/mesonbuild/libffi/pull/1/files#diff-969b60ad3d206fd45c208e266ccfed38, but unmerged.

btw, it appears that the automated windows build via docker hub is still not possible <sup>[[1]](https://beta.docs.docker.com/v17.12/docker-hub/builds/#limitations) [[2]](https://forums.docker.com/t/automated-windows-builds/24192/8)</sup>, are there plans to consider some alternative hook?

Snaipe:
I've opted to use the `cirrusci/windowsservercore:cmake` image for now, and install python and meson on top of it during each build. It's not great, but it works reliably in the absence of automated image builds for a custom image with meson preinstalled. (you can see the [current cirrus ci setup over on BoxFort](https://github.com/Snaipe/BoxFort/blob/master/.cirrus.yml) -- it's very likely that I'll adopt a very similar config for this repo).

It also looks like libffi has already been wrapped by the gstreamer project, so it's very likely that I'll use that if it builds on all target platforms.

Snaipe:
This was included in #319

PR #319, opened 2019-12-29
wip: ci: switch to CirrusCI.

Snaipe:
I've been experimenting for a while with Cirrus CI on BoxFort after seeing @am11's pull request. Cirrus, along with meson, has done wonders for the project and actively reduces my maintenance pains.

I will have this supersede #287.

I'm also including here the switch from dyncall to libffi -- that's because I don't want to spend too much time on making dyncall work with the CI environment. If I can get libffi to work on every task defined in .cirrus.yml, I will deem the switch safe to do.


.travis.yml
```diff
@@ -1,185 +0,0 @@
-language: c
```
am11: probably, appveyor.yml can also be removed in the same vein. 😄

Snaipe: LOL, turns out that the appveyor.yaml got [nuked by mistake](https://github.com/Snaipe/Criterion/commit/017b3003f8bbb452502f29ab3981855191df7581) when I merged the meson branch. Way to be proactive!

.cirrus.yml
```diff
@@ -0,0 +1,78 @@
...
+Windows_task:
+  windows_container:
+    image: cirrusci/windowsservercore:cmake
+    os_version: 2019
+  env:
+    PATH: C:\\Python;C:\\Python\\Scripts;C:\\ProgramData\\chocolatey\\lib\\ninja\\tools;%PATH%
+  setup_script: |-
+    choco install -y --no-progress python3 --params "/InstallDir:C:\Python"
+    choco install -y --no-progress ninja cmake
+    pip install meson cram==0.7
```
am11: meson setup seems to be consume some minutes in BoxFort: https://cirrus-ci.com/task/5856113753063424,
maybe it could be added to the [cache](https://cirrus-ci.org/guide/writing-tasks/#cache-instruction) (with sane fingerprint).

.travis.yml
am11: Great! they can also be nuked from https://github.com/Snaipe/Criterion/settings/installations and https://github.com/Snaipe/Criterion/settings/hooks to hide them from CI run in PRs.

.cirrus.yml
Snaipe: Yeah, I hear you. I've had mixed results with getting the cache to somewhere that didn't break previous builds... one of the most problematic aspects was when I left my cirrus-ci WIP sit for a few weeks alone, and then re-triggering a build, only to find out that I got an older dependency cached while the underlying image moved forward.

For the scope of this PR, I think I'll leave the cache alone, and just make this "slow" but correct. We can optimize build times further when we get back to a stable build and CI environment.

.cirrus.yml
```diff
@@ -0,0 +1,78 @@
...
+    brew install meson cmake pkg-config libffi libgit2
+    pip3 install cram==0.7
+  <<: *pipeline
+
+FreeBSD_task:
+  freebsd_instance:
+    image_family: freebsd-12-0
+  setup_script: |-
+    pkg install -y py36-pip meson git libffi libgit2
+    pip-3.6 install cram==0.7
```
am11: empirically speaking) In FreeBSD, maintainers mercilessly deprecate and remove the packages from ports quite often than other packaging managements. Would probably make sense to keep pip version-less, or use the one from python [e.g.](https://github.com/Snaipe/Criterion/pull/287/files#diff-0f7067c31caedf7df2739c9b07efa122R31):

```diff
- pkg install -y py36-pip meson git libffi libgit2
- pip-3.6 install cram==0.7
+ pkg install -y cmake python meson git libffi libgit2
+ python -m ensurepip && python -m pip install --upgrade pip
+ python -m pip install --user cram==0.7
```

Snaipe: Do they change these kind of versions & package names within a major FreeBSD release? (as in, is using freebsd-12-0 ensuring that py36-pip is there?)

In any case, I think I'll take your word for it here, especially if I want to easily upgrade images in the future.

am11: Yup, CI-like use-cases are not very FreeBSD-friendly. For example, I was pulling in a specific version of node.js in CI job, and suddenly (like hours later) someone just ninja'd in a newer version and "removed" the old one. Full story: https://github.com/cbsd/cbsd/issues/367#issuecomment-459356238. I ended up [scraping their webpage](https://github.com/sass/node-sass/blob/0aa9518068ed969300dfb3e6925449bbbff49c00/.cirrus.yml#L43-L45) for "whatever available version". 😅

ps - there are just too many package management systems, all with different/contradicting philosophies

Snaipe:
This seems to be going well. On the last remaining failures (FreeBSD, Windows, Linux on ARM), these tasks managed to compile Criterion and its dependencies, but then got stuck at linking or testing.

* Windows can't seem to properly link its own dependencies. Not sure what's causing this, and the logs are not displaying the actual link command, so it's hard to evaluate.
* FreeBSD can't find libintl, and thus can't resolve libintl_gettext(). This is troubling, as libintl.h was evidently found and used, but the library somehow wasn't found on the system. I think it's a problem with pkg-config.
* The ARM builds seem to fail at the test phase, and for good reasons: the exception catching is completely broken. I have no idea if this is something that happens on real hardware or just qemu.

Snaipe:
I've dropped the Arm & AArch64 jobs for this PR. I just want feature parity with travis & appveyor, and the FreeBSD task is a freebie since getting it to work was a piece of cake.

On the other hand the Arm failures are nontrivial. I'll address them in another PR.

Snaipe:
Alright! I got everything to pass.

There's a handful of annoying things that remain:

* FreeBSD, MacOS, and Windows all build without locale support. While the latter two are not that surprising, I'm a bit puzzled regarding why meson can't find libintf on FreeBSD.
* The Windows build uses MinGW GCC. We also need a proper MSVC build on both BoxFort and Criterion.
* The Windows MinGW build doesn't run the system (cram) tests. That's because cram needs sh to be present on the system. I think this can be fixed easily for the MinGW task as the container image has git installed, which means it should also be shipping a functional bash, and I suspect we could adjust PATH.

Unfortunately, I can't address the last two points now as I don't have access to a Windows, and still won't have for the next 2-3 weeks. @am11 if you're game and have some time to spare, maybe you could have a look at figuring that out?

Anyhow, I'm going to clean this up a bit, merge it, and open a few issues to for the above annoyances.

PR #320, opened 2019-12-30
ci: add ARM and AArch64 builds

Snaipe:
As explained in #319, I left out the ARM builds because there were non-trivial failures.

To be more precise, exception handling is plainly broken -- criterion doesn't catch anything at all.

Let's figure out why and fix that in this PR.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -32,6 +32,22 @@ Alpine (gcc,i386)_task:
   <<: *pipeline
   test_script: ninja -C build test
 
+Alpine (gcc,arm)_task:
+  container:
+    image: snaipe/ci-meson:alpine-armhf
+  env:
+    CIRRUS_SHELL: /sbin/qemu-sh
+  setup_script: *alpine-deps
+  <<: *pipeline
+
+Alpine (gcc,aarch64)_task:
+  container:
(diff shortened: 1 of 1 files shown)


am11:
btw, github also has introduced ci under the overarching concept [github-actions](https://help.github.com/en/actions). there is also a [self-hosted runners](https://github.blog/2019-11-05-self-hosted-runners-for-github-actions-is-now-in-beta/) options. [currently sporting](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idruns-on) macos, ubuntu and windows with arm{32,64} and x{64,86} support. no *bsd support yet - [example](https://github.com/rust-lang/stdarch/blob/master/.github/workflows/main.yml) which shows up under [<kbd>▷ Actions</kbd> tab](https://github.com/rust-lang/stdarch/actions).

Snaipe:
Thanks to @MrAnno's efforts, this now works out of the box. It's slow as hell since it runs with qemu-user, and does not catch issues that would otherwise appear on real hardware, but it's something.

In the time that elapsed since I first opened this PR, Cirrus does seem to have added support for arm64 containers, so I'll leave the effort to switch to them as a later improvement. This is good enough for now.

PR #324, opened 2019-12-31
api: use the correct printf specifiers on mingw-gcc.

### LINKED ISSUE #322: Fix warnings on MinGW build
Compiling the samples on MinGW yields a flurry of warnings: https://cirrus-ci.com/task/4808504322031616

This is likely because the printf format specifiers are different between the windows crt and glibc.

Snaipe:
MinGW uses by default the windows CRT for printf, but can be convinced
to use its own ANSI-compatible printf implementation by defining the
__USE_MINGW_ANSI_STDIO macro.

This fix is two-fold: first, we use the "I64" specifier prefix instead of "ll"
for the long long type modifier if we're using MSVC.

Then, since cr_asprintf uses printf and is compiled in libcriterion.so,
we have to make a choice on which printf to use for our users.
Naturally, we prefer the standards-compliant one.

Fixes #322.

### DIFF
--- a/include/criterion/internal/asprintf-compat.h
+++ b/include/criterion/internal/asprintf-compat.h
@@ -34,9 +34,22 @@
 
 #include "common.h"
 
+/* MinGW makes the distinction between the printf and gnu_printf archetypes,
+   because they provide a C99-compliant implementation of printf when
+   the __USE_MINGW_ANSI_STDIO macro is defined. Since cr_asprintf calls
+   the vsnprintf function and is compiled in libcriterion, we need to make
+   the choice of which printf function to use.
+
+   Naturally, we use the C99-compliant implementation. */
+#ifdef __MINGW32__
+# define CRI_PRINTF_FORMAT gnu_printf
+#else
+# define CRI_PRINTF_FORMAT printf
+#endif
+
 CR_BEGIN_C_API
 
-CR_FORMAT(printf, 2, 3)
+CR_FORMAT(CRI_PRINTF_FORMAT, 2, 3)
 CR_API int cr_asprintf(char **strp, const char *fmt, ...);
 CR_API int cr_vasprintf(char **strp, const char *fmt, va_list ap);
 CR_API void cr_asprintf_free(char *buf);
--- a/include/criterion/internal/assert/op.h
+++ b/include/criterion/internal/assert/op.h
@@ -353,8 +353,6 @@
         }                                                                   \
     } while (0)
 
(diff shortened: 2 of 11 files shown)


am11:
Was there any problem using %zu with msvc? When I tested %zu, it works just fine with the versions of VS Criterion project supports. It has been a while c99 (and c11) support was incorporated in vcr.

Snaipe:
I think there was when I was first developing that part of the code, which means with MSVC 14 and the associated crt. I don't know if this is still a version we want to support, but my gut feeling is telling me yes for now.

am11:
Ah right, sorry i was thinking we support 2017 onwards (15). I have tested with 2017 and 2019 (16).

The acquisition of VS build tools (gui-less) and VS community edition these days is very simple, and VS2019 supports all versions of Windows including XP (rel. 2001). So imo, dropping 2015 would not be too bad if it shaves off some #preprocessor branches in code base (although its extended support ends till 2015).

ps: current master is not compiling with msvc, i am fighting with libffi build (preparing PR for upstream to have a CMakeLists.txt file) 😉

Snaipe:
Oh. I should have guessed libffi would be troublesome to build on MSVC.

The libffi wrap that we're using has meson definitions, maybe these can be fixed rather than trying to get a CMakeLists.txt to upstream? I'm seeing a lot of libffi ports that add cmake support, but none of that is upstream, so I'm guessing that they want to keep using autotools.

Snaipe:
On the topic of shaving preprocessor branches, I think the simplest thing to do here, since we only expose these format specifiers through cr_asprintf which in turn uses whatever vsnprintf implementation is available, is to add a configure-time check for the "ll" and "z" length modifiers, as well as what PRI{u,d}64 means on the current platform. We can then define the right format specifiers in a config.h installed in criterion/.

I don't think I'll do that in this PR, but I'll open an issue to remember to do that later.

Snaipe:
There are no new warnings with the mingw-gcc build, so this seems good to merge.

am11:
> I'm seeing a lot of libffi ports that add cmake support, but none of that is upstream, so I'm guessing that they want to keep using autotools.

The maintainers seem to be OK to have cmake configs for windows, so i prepared a multi-platform patch: https://github.com/libffi/libffi/issues/313#issuecomment-571447391. 🤞

Thing with libffi meson ports is that it is totally isolated and apparently playing catch-up for sometime and still missing some introspection code.. so i guess if we cut down this middle wrapper and consume straight from the source (like we have for other deps), that would bring about simplicity.

PR #328, opened 2020-01-02
ci: Fixed paths to isdir.py and dependencies when building as meson subproject

### LINKED ISSUE #327: Cannot include Criterion as meson subproject
Hello!

I've tried to use Criterion in my project as meson's wrap-based subproject, and during simplest `meson build` I got some issues related build scripts using wrong paths. `meson.source_root()` is used for `ci/isdir.py` and `dependencies`, and in a subproject there are unfindable by default:

```
subprojects/criterion/meson.build:20:0: ERROR: Program or command '/home/tomasz/criterion-bug-source-dir/ci/isdir.py' not found or not executable
subprojects/boxfort/meson.build:19:0: ERROR: Program or command '/home/tomasz/criterion-bug-source-dir/ci/isdir.py' not found or not executable
```

Thankfully using `meson.current_source_dir()` is enough - see the below ugly fix. I'll try to produce a pull request for that. :)

Here's the minimal repo I've been able to reproduce this, with the ugly in-place fixes: https://hg.sr.ht/~tsieprawski/criterion-bug-source-dir/

```bash
sed -i "s|source_root|current_source_dir|" subprojects/criterion/meson.build
sed -i "s|'\.\.\/dependencies\/nanopb|meson.current_source_dir() + '/../dependencies/nanopb|" subprojects/criterion/src/meson.build
sed -i "s|source_root|current_source_dir|" subprojects/boxfort/meson.build
```

And the `asciinema` video: [video.cast.zip](https://github.com/Snaipe/Criterion/files/4016478/video.cast.zip)

tsieprawski:
Fixes #327

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -57,9 +57,9 @@ sources = files(
 	'string/i18n.c',
 	'string/xxd.c',
 
-	'../dependencies/nanopb/pb_common.c',
-	'../dependencies/nanopb/pb_encode.c',
-	'../dependencies/nanopb/pb_decode.c',
+	meson.current_source_dir() + '/../dependencies/nanopb/pb_common.c',
+	meson.current_source_dir() + '/../dependencies/nanopb/pb_encode.c',
+	meson.current_source_dir() + '/../dependencies/nanopb/pb_decode.c',
 )
 
 configure_file(input         : 'config.h.in',
--- a/meson.build
+++ b/meson.build
@@ -9,7 +9,7 @@ prefix    = get_option('prefix')
 localedir = get_option('localedir')
 
 # Helper scripts
-auxdir  = join_paths(meson.source_root(), 'ci')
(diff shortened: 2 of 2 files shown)


src/meson.build
```diff
@@ -57,9 +57,9 @@ sources = files(
 	'string/i18n.c',
 	'string/xxd.c',
 
-	'../dependencies/nanopb/pb_common.c',
-	'../dependencies/nanopb/pb_encode.c',
-	'../dependencies/nanopb/pb_decode.c',
+	meson.current_source_dir() + '/../dependencies/nanopb/pb_common.c',
```
Snaipe: Mmh, yeah, maybe it'll be better to move these types of libs under ./src, or make them full blown deps. Nanopb recommends compiling these files directly into the project, but I think I'd rather build a static lib.

For now, I'm fine with this coming in as-is.

Issue #332, opened 2020-01-06
Package Criterion on Debian


Snaipe:
This issue tracks everything that needs to be done in order to get Criterion packaged on Debian.

* [ ] Upgrade generated protobuf sources to use nanopb 0.4.0.
* [ ] Always re-generate protobuf sources during build if the nanopb generator is available locally.
* [ ] Move nanopb from a submodule to a meson wrap, and use the system nanopb if available.

Another big one is that I think that I want to release 2.4.0 first -- or at least, verify that the new assertion API did not break anything badly and that everything is OK and stable. We're quite close from a release anyway, so it shouldn't delay things by much. I'll draft a roadmap towards the release soon.

Snaipe:
I believe Criterion has been packaged on Debian; closing this

PR #333, opened 2020-01-06
Update nanopb and link against lib.

Snaipe:
This partially addresses #332


Snaipe:
Whoa. What's up with these runtimes on FreeBSD? Either the new nanopb is slow as hell, or something is very wrong.

Snaipe:
Okay, so we either have a race condition somewhere in the FreeBSD build, or we suffer from bad performance on the cirrus CI hosts somehow. I think it's a separate issue though, so I think this is good to go.

PR #334, opened 2020-01-07
meson: Add options to not-build samples and tests.

tsieprawski:
tl;dr same as https://github.com/Snaipe/BoxFort/pull/27 but for Criterion.

When using Criterion as a subproject, meson test runs Criiterion's tests. I believe a typical developer developing their super project wants to run only own tests, not dependencies' ones.

This PR adds -Dsamples and -Dtests options to enable/disable, like this:

```sh
meson setup -Dsamples=false -Dtests=false build && cd build && meson test
# Or as a subproject
meson setup -Dcriterion:samples=false -Dcriterion:tests=false build && cd build && meson test
```

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -225,5 +225,11 @@ deps += [
 
 subdir('include')
 subdir('src')
-subdir('samples')
-subdir('test')
+
+if get_option('samples')
+	subdir('samples')
+endif
+
+if get_option('tests')
+	subdir('test')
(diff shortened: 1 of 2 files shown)


Snaipe:
LGTM, thanks for the fix! I also find it pretty funny that I added a `tests` option but didn't do anything with it. Oh well :)

PR #343, opened 2020-02-25
meson: fix 'theories' option to be actually disablable

tsieprawski:
With `meson -Dtheories=disabled` not only `libffi` is not used as a dependency, but also `src/core/theories.c` and theories samples and tests are not being built.

My use case:
- I try to build a minimal `Dockerfile` for my CI-ing my software. I don't need `ffi`;
- I use Criterion for tests :) ;
- I don't use theories.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@
 *.sw[op]
 .*.sw[op]
 build
+subprojects/boxfort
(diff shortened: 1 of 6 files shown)


Snaipe:
Merged as 5e7aff2.

PR #344, opened 2020-02-25
meson: add option to disable libgit2

tsieprawski:
Added the meson option to disable dependency of `libgit2`.

My use case is:
- I try to produce a minimal `Dockerfile` for my software and I am on very early stages. In the end it will be used in CI;
- I use Criterion :) ;
- `libgit2` is a big library which adds both to `docker build` time and taking space, and all it brings is to diff strings in 2 places (`src/core/assert.c` + `src/log/normal.c` via `cri_diff_buffer_to_buffer`).
- Right now it is good enough for me to not-have beautiful git-diffs. I still get the failed assertion anyway and I can get proper git-diff running the test locally. :)

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -181,7 +181,7 @@ if not nanopb.found()
 endif
 
 libgit2 = dependency('libgit2', required: false)
-if not libgit2.found()
+if not libgit2.found() and get_option('diffs').enabled()
 	libgit2_proj = cmake.subproject('libgit2',
 		cmake_options: [
 			'-DBUILD_SHARED_LIBS=OFF',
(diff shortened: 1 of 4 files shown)


Snaipe:
Thanks for the patch. A couple of comments:

* Can you have a look at the failing builds? I think this breaks tests and no longer prints any diffs.
* Rather than huge ifdefs, I'd prefer if diff.c was compiled-in only when libgit2 is found, and otherwise a diff_stubs.c file with your empty stubs would get used otherwise.
* Can the option be named more after enabling the "diff feature" rather than "libgit2"? maybe "assert-diffs" or just "diffs".

tsieprawski:
Sorry, life got into the way and I'm constantly forgetting to get back to this PR.
`diff_stubs.c` and option `diffs` are cool, I'll go that way.

tsieprawski:
There are 3 build failures. FreeBSD and Windows seem to be not-mine issues, MacOS are crashes in boxfort tests - here I am not 100% sure. Can you help @Snaipe ?

Snaipe:
Seems unrelated. Other than that, LGTM.

PR #361, opened 2020-10-04
xml: Fix escape of special characters

Eazhi:
The current xml writer produces incorrect xml files when using `cr_assert_lt` or `cr_assert_gt`. Special characters like `>`, `<`, `&` were not escaped.

Enclosing the error messages in a CDATA section allows us to not have to escape such characters and keep the XML human readable.


Snaipe:
Thanks for the fix!

PR #366, opened 2020-12-05
Simplify autotools example

Ambroisie:
I have been helping students setup a big project using Autotools and Criterion at my school. Having used both of them together before, I was surprised to see that your example setup is way more complicated than what I had arrived at by reading the documentation.

I have both simplified the setup to use environment variables to setup TAP output (making the use of a wrapper shell script redundant), and made the example stricter in terms of good practices regarding the use of Autotools.

### DIFF
--- a/dev/autotools/autogen.sh
+++ b/dev/autotools/autogen.sh
@@ -1,18 +1,3 @@
 #!/bin/sh -e
 
-# copy TAP driver into build-aux
-automake_ver=$(automake --version | \grep -E -o '[0-9]\.[0-9]{2}')
-automake_dir=$(dirname $(which automake))/..
-
-mkdir -p build-aux
-cp -f $automake_dir/share/automake-$automake_ver/tap-driver.sh build-aux
-
-# create criterion TAP log compiler
-# this is necessary to print TAP (and only TAP) on the standard output,
-# and always exit with 0 to let the TAP driver handle errors itself.
-echo >build-aux/criterion-tap-test """#!/bin/sh
-\$1 -Otap:- --always-succeed
-"""
-chmod +x build-aux/criterion-tap-test
-
 autoreconf -vi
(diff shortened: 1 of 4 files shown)


Ambroisie:
I just noticed that I forgot to update the README, will update soon.

Ambroisie:
I updated the README in accordance with my changes.

Ambroisie:
One last thing would be to add some logic to handle the fact that `Criterion` is *not* present on the host computer... Logically a missing test-dependency should not break the build, but simply disable the tests using it.

dev/autotools/autogen.sh
```diff
@@ -1,18 +1,3 @@
 #!/bin/sh -e
 
-# copy TAP driver into build-aux
-automake_ver=$(automake --version | \grep -E -o '[0-9]\.[0-9]{2}')
-automake_dir=$(dirname $(which automake))/..
-
-mkdir -p build-aux
-cp -f $automake_dir/share/automake-$automake_ver/tap-driver.sh build-aux
```
Snaipe, line 8: Isn't this still required? I remember autotools not putting the tap driver in the aux directory, which made the whole thing fail.

Snaipe (commented):
This looks good, I just have one question/nit before merging this.

dev/autotools/autogen.sh
Ambroisie: I don't recall having any issues, the AUX_DIR macro is supposed to instruct Autotools to put its files in that directory IIRC. 

I'll try the setup from scratch once I come back to my computer and let you know.

Ambroisie: It works on my machine (using Arch linux).

Snaipe: Thanks for confirming. I'll merge your changes as-is, thank you for the fix!

PR #369, opened 2021-01-08
Fix C++20 compatibility: Removed feature used

Ersikan:
The `std::allocator<T>` type in the C++ standard library does not declare
the `const_pointer` associated type anymore, as per the update of the
C++ standard.

This fix replaces its usage with a normal constant pointer:

```diff
-typename std::allocator<void>::const_pointer
+const std::allocator<void>::value_type *
```


Snaipe:
Thanks for the fix!

PR #370, opened 2021-01-12
log: Fix crash in `cr_log_info()` on empty string

### LINKED ISSUE #311: Segmentation fault when log message starts with an empty string
When a log message starts with an empty string the test crashes with a segmantation fault (only tested on macOS High Sierra). 

All of the log messages in the following test would cause a segmentation fault:
```c
Test(criterion, segmentation_fault_when_log_message_starts_with_empty_string)
{
    cr_log_info("");
    cr_log_info("%s", "");
    cr_log_info("%s characters after formatted empty string", "");
}
```

When the message is not starting with an empty string the message is printed as expected:
```c
Test(criterion, log_no_segmentation_fault)
{
    cr_log_info("characters in front of formatted empty string %s", "");
}
```
Thank you for this great tool!

Ersikan:
Fixes #311.

The issue was caused by an unchecked dereferencing of the first token
returned by `strtok_r()`, which is NULL if the message is empty.

### DIFF
--- a/src/log/normal.c
+++ b/src/log/normal.c
@@ -427,6 +427,9 @@ void normal_log_test_abort(CR_UNUSED struct criterion_test_stats *stats, const c
 
 void normal_log_message(enum criterion_severity severity, const char *msg)
 {
+    if (*msg == '\0')
+        return;
+
     char *dup       = strdup(msg);
     char *saveptr   = NULL;
     char *line      = strtok_r(dup, "\n", &saveptr);


src/log/normal.c
```diff
@@ -431,10 +431,12 @@ void normal_log_message(enum criterion_severity severity, const char *msg)
...
-    do {
-        if (*line != '\0')
-            criterion_log_noformat(severity, line);
-    } while ((line = strtok_r(NULL, "\n", &saveptr)));
+    if (line) {
+        do {
+            if (*line != '\0')
+                criterion_log_noformat(severity, line);
+        } while ((line = strtok_r(NULL, "\n", &saveptr)));
+    }
```
Snaipe, line 439: I think I'd prefer if we just short-circuited the logic at the very beginning of normal_log_message -- that way, if we have an empty string, we don't bother going through the malloc/copy/do-nothing/free cycle.

Ersikan:
@Snaipe I updated the fix. Now the function does an early return if the message is empty, bypassing the allocation and deallocation.

PR #403, opened 2021-12-30
Fix compiler warnings

MrAnno:


Snaipe:
Huh, I would have thought that something else would complain since void (*)(void *) isn't compatible with void (*)(struct criterion_test *), but I guess we're still calling the function with a pointer which ends up being implicitly converted to void *. I think that's UB, but I don't think it matters much in practice. I'd like to replace report hooks with something else down the line, so safety issues can be solved then.

Snaipe:
LGTM.

PR #404, opened 2021-12-31
Fix UB, support ARM/Thumb interworking, and run ARM tests in CI

### LINKED ISSUE #347: Performance issue with unusually high "max open files" ulimit
Hello,

don't ask me why (or where it's configured) but our Jenkins server is configured for 1M max open files, as per /proc/<pidof jenkins>/limits.

I see very poor runtime of tests, multiple orders of magnitudes longer runtimes for some tests, compored to running our tests locally on my workstation (from <1s to >30s)

I found that the open files ulimit is the main issue. When I wrap the tests with a script that does "ulimit -n 256", the runtime of the affected tests goes back to normal. I found this in strace traces, where I saw that fcntl() is called for every possible file descriptor. The code is in BoxFort, specifically this loop: https://github.com/Snaipe/BoxFort/blob/master/src/sandbox-posix.c#L508

What's curious: some tests do not show the poor performance. Even if I artificially compile the exact same code in two tests, one is quick even inside Jenkins but the other one slow as hell.

Do you think the loop can be smarter to handle such systems?

MrAnno:
Sorry, this was my fault. When I fixed the Windows pipeline, I introduced some boilerplate at the same time.
I have refactored the pipeline, so this won't happen again.

I went ahead and started using Cirrus CI's arm containers too, they are much faster.

Depends on Snaipe/BoxFort#35
Fixes #347

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -44,6 +44,12 @@ Alpine (gcc,arm32v6)_task:
     pip3 install cram==0.7
   <<: *pipeline
 
+Alpine (gcc,arm32v7)_task:
+  arm_container:
+    image: arm32v7/alpine:3
+  setup_script: *alpine-arm-deps
+  <<: *pipeline
+
 Alpine (gcc,aarch64)_task:
   arm_container:
     image: arm64v8/alpine:3
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,10 +1,17 @@
-_pipeline: &pipeline
-  info_script: |
-    meson --version
-    ninja --version
+.compile: &compile
   submodule-update_script: git submodule update --init --recursive
   configure_script: meson -Db_lundef=false build
   compile_script: ninja -C build
+
+.test: &test
+  test_script: ninja -C build test
+
+.pipeline: &pipeline
+  info_script: |
+    meson --version
+    ninja --version
+  <<: *compile
+  <<: *test
 
 Debian (gcc)_task:
   container:
@@ -14,7 +21,6 @@ Debian (gcc)_task:
     apt-get install -y pkg-config cmake libgit2-dev libffi-dev libnanomsg-dev
     pip3 install cram==0.7
   <<: *pipeline
-  test_script: ninja -C build test
 
 Alpine (gcc,x86_64)_task:
   container:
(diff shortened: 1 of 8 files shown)


MrAnno:
arm32v7 seems to be completely broken, I'll have a deeper look at it soon.
arm32v6 has some exception handling issues.
aarch64 works properly, some float precision issues needs to be fixed in tests.

Snaipe:
Looking at the arm64 failure, we should fix the test to not be as picky; I don't think we care much whether this prints 0.33333333333333 or 0.33333333333317 or whatever else, testing the floating-point precision of the target platform is not what this is about. Let's change the pattern to `0\.3.*`.

MrAnno:
Exception handling: `CR_TEST_PROTOTYPE_` is marked as `extern C`. It is undefined behavior to throw exceptions in such a function. `-fexceptions` will take care of this issue, but I would like to take some time, because I think a cleaner resolution might exist.

MrAnno:
arm32v6 and arm64v8 are finally green. I'll investigate and hopefully fix arm32v7 tomorrow.

Happy New Year!

Snaipe:
Happy New Year! :tada: 

The arm v7 failures aren't looking too good given that the BoxFort tests are getting SIGABRT'ed. I'd wager it's yet another trampoline issue or something with main() not being found, though I obviously haven't investigated in detail. Though, this seems to be good progress at least.

MrAnno:
I think we have run into ARM's Thumb mode, because the address of main is off by one:
https://developer.arm.com/documentation/dui0471/m/interworking-arm-and-thumb/pointers-to-functions-in-thumb-state

Compiling with `-marm` would solve that for us safely, but alternatively we could trick that one offset back manually.

> Thumb code is able to provide up to 65% of the code size of Arm, and 160% of the performance of an equivalent Arm processor connected to a 16-bit memory.
> For critical code size applications, the alternative 16-bit Thumb Mode reduces code by more than 30% with a minimal performance penalty.

MrAnno:
"Providing your own main" might become an issue if we end up using `-marm`, so I think it would be better to fix the offset itself. I'm doing some tests.

Snaipe:
I think -marm might just be a stopgap since we'd just be compiling criterion as non-thumb, and it wouldn't help users that want to use the library in thumb mode.

I'm not sure what the proper way to detect thumb mode would be, but I'd be fine with just checking if the LSB is set if the architecture is ARM.

MrAnno:
BoxFort revision has been updated.

MrAnno:
I haven't had this much fun for a long time, especially not with ASM and self-modifying code.

I can only repeat myself: Criterion, BoxFort, and Mimick are amazing, they are pieces of art. :)

Snaipe:
> I haven't had this much fun for a long time, especially not with ASM and self-modifying code.
>
> I can only repeat myself: Criterion, BoxFort, and Mimick are amazing, they are pieces of art. :)

It's all about making the most out of the software landscape, the more frustrating to use everything is, the more innovation can be squeezed out! That aside, it's true that these three projects have been the most challenging, brain-racking, and fun to write.

The PR looks good to me; thanks again for the fixes.

PR #409, opened 2022-01-01
boxfort: fallback to /proc/self/fd

### LINKED ISSUE #347: Performance issue with unusually high "max open files" ulimit

Snaipe:
Fixes #347.

### DIFF
--- a/subprojects/boxfort.wrap
+++ b/subprojects/boxfort.wrap
@@ -1,4 +1,4 @@
 [wrap-git]
 directory = boxfort
 url = https://github.com/Snaipe/BoxFort.git
-revision = cec0db7ab13c51d0baa01e2680b0edee2f7d317a
+revision = c91f88b1476509a8c7472ca712130acab49e3cc1


Snaipe:
Superseded by #404.

PR #410, opened 2022-01-03
misc,cram: fix cram patch not working from ninja test

Snaipe:
This restores the ability to patch cram failures via `CRAM=-iy ninja test`,
which broke when switching to meson: because the test directory passed
to cram was an absolute path, the patch program would also get fed absolute
path, which in turn declined to do anything for safety reasons.

This commit forces the cram test to execute with the working directory
set to the project root, passing `test/cram` as path instead.
Interestingly, changing the workdir to test/cram and then using `.` as
test path did not work.

### DIFF
--- a/test/cram/meson.build
+++ b/test/cram/meson.build
@@ -18,6 +18,7 @@ if sh.found() and cram.found()
 
 	test('cram tests', cram,
 			timeout: 240,
-			args: ['--shell=' + sh.path(), meson.current_source_dir()],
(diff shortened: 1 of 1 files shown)


MrAnno:
The new native ARM containers in #404 have meson v0.59, so `project_source_root()` (> v0.56) won't be a problem.

Snaipe:
Yeah, and it doesn't look like I can easily try/catch the existence of the function to fallback to something else. I'm in no mood to rebuild all of the container images, so I'll just use a hack with current_source_dir instead.

PR #413, opened 2022-01-04
Fix build with older gcc versions

MrAnno:
This is a fix for something that was introduced in #405, so no changelog entry is needed.

@Snaipe
The PR is opened against `bleeding`, but we are in the middle of a release. Which workflow do you prefer? Should I change the base to `master` and then merge that branch back into `bleeding`?

### DIFF
--- a/test/unit/asprintf.c
+++ b/test/unit/asprintf.c
@@ -17,11 +17,11 @@
 
 Test(asprintf, valid) {
     assert_cr_asprintf_format_is_valid("%c", 'a');
-    assert_cr_asprintf_format_is_valid("%hd", 42);
+    assert_cr_asprintf_format_is_valid("%hd", (short) 42);
     assert_cr_asprintf_format_is_valid("%d", 42);
     assert_cr_asprintf_format_is_valid("%ld", 42L);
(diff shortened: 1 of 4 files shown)


test/unit/asprintf.c
```diff
@@ -17,11 +17,11 @@
 
 Test(asprintf, valid) {
     assert_cr_asprintf_format_is_valid("%c", 'a');
-    assert_cr_asprintf_format_is_valid("%hd", 42);
+    assert_cr_asprintf_format_is_valid("%hd", (short) 42);
```
Snaipe, line 20: I'm surprised gcc even cares given that any value shorter than int gets implicitly passed as an int through varargs. I guess the warning is there just for correctness.

Snaipe (approved):
LGTM.

test/unit/asprintf.c
MrAnno: I was surprised too. :)

Snaipe:
Is this OK to merge? Feel free to do so yourself if you're happy with your changes, since they've already got my approval and haven't since changed.

MrAnno:
I think it is ready. Thanks for your review.

PR #414, opened 2022-01-05
Automated release workflow

MrAnno:
This PR tries to automate the last few steps of the Criterion release process.

When a commit is pushed to the master branch or a version tag is created, the release workflow starts automatically.
The workflow can be restarted manually on either a tag or a branch if needed.

The release draft workflow creates the source tarball (which includes all subprojects), builds Criterion from the tarball, and creates binary packages.

c2

Binary packages are created within a CentOS 7 container. This sounds terrible, but it ensures that Criterion is built with an older glibc version (v2.17) so that it will work on a wide variety of Linux distros.

The source tarball and the binaries then will be attached to the release on the Releases page.

In case new commits are pushed to the master branch, a pre-release will be published automatically using `git describe --long` for the name of the pre-release (this creates unsigned lightweight tags).

c4

When the workflow is triggered by a new version tag (vX.Y.Z), a final release DRAFT will be made. This needs to be reviewed and published manually.

c1

The workflow verifies the release tag as well, it opens an issue to warn maintainers about unsigned or lightweight release tags.

c3

----

Another workflow tries to fast-forward bleeding. If that's not possible, an issue will be raised.

### DIFF
--- a/.github/workflows/ff-bleeding.yml
+++ b/.github/workflows/ff-bleeding.yml
@@ -1,0 +1,39 @@
+name: Fast-forward bleeding
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+
+env:
+  FF_ISSUE_TITLE: "No-FF: bleeding is behind master"
+  GITHUB_TOKEN: ${{ github.token }}
+
+jobs:
+  fast-forward-bleeding:
+    name: Fast-forward bleeding
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out Criterion
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Configure git
+        run: |
+          git config user.name "github-actions"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+
+      - name: Fast-forward bleeding
+        run: |
+          git checkout bleeding
+          if ! git merge --ff-only --no-edit master; then
+            issue_opened="$(gh issue list --state open --jq ".[] | select(.title == \"${FF_ISSUE_TITLE}\")" --json title)"
+            if [ -z "${issue_opened}" ]; then
+              gh issue create --title "${FF_ISSUE_TITLE}" --body ""
+            fi
+            exit 1
+          fi
+
+          git push origin bleeding
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,163 @@
+name: Release
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+    tags:
+      - v[0-9]+.[0-9]+.[0-9]+
+
+env:
+  RELEASE_BRANCH: master
+  RELEASE_CANDIDATE_TAG_PATTERN: ^v[0-9]+.[0-9]+.[0-9]+-rc
+  RELEASE_SIGNING_KEY_ID: 75B0 21D9 64AD 8693 24CD  B74D DDD7 BCCD 44BB 9A9D
+  GITHUB_TOKEN: ${{ github.token }}
+
+jobs:
+  release-tag-verification:
+    name: "Release tag verification"
+    if: github.ref_type == 'tag'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out Criterion
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.ref }}
+          fetch-depth: 0
+      - name: Release tag verification
+        run: |
+          gpg --keyserver hkps://keyserver.ubuntu.com --recv-keys "${RELEASE_SIGNING_KEY_ID}"
+          if ! git tag --verify "$GITHUB_REF_NAME" > git-verify-tag.log 2>&1; then
+            gh issue create --title "$GITHUB_REF_NAME tag verification failure" --body-file git-verify-tag.log
+            exit 1
+          fi
+
+  tarball:
+    name: "Source tarball"
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.detect-version.outputs.version }}
+
+    steps:
+      - name: Check out Criterion
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Configure git
+        run: |
+          git config user.name "github-actions"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+
+      - name: Install dist dependencies
+        run: sudo apt-get update && sudo apt-get install -y meson
+
+      - name: Detect version
+        id: detect-version
+        run: |
+          if [[ "$GITHUB_REF_TYPE" == "tag" ]]; then
+            version="$GITHUB_REF_NAME"
+          else
+            version="$(git describe --long)"
+            if [[ ! "${version}" =~ ${RELEASE_CANDIDATE_TAG_PATTERN} ]]; then
+              printf "Pre-release requires an annotated tag with '-rc' suffix (actual: %s)\n" "${version}"
+              exit 1
+            fi
+          fi
+          version="${version:1}"
+          echo "${version}"
+          echo "::set-output name=version::${version}"
+
+      - name: Create source tarball
+        env:
+          version: ${{ steps.detect-version.outputs.version }}
+        run: |
+          meson rewrite kwargs set project / version "${version}"
+          git commit --allow-empty -a -m "release: version bump to v${version}"
+          meson setup -Dc_std=gnu11 --wrap-mode forcefallback build
+          meson dist -C build --include-subprojects --formats xztar
+
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: source-tarball
+          path: build/meson-dist/criterion-*.tar.xz
+          if-no-files-found: error
+
+  binaries:
+    name: "Binaries"
+    runs-on: ubuntu-latest
+    # glibc v2.17 builder
+    container: centos:7
+    needs: [tarball]
+    env:
+      version: ${{ needs.tarball.outputs.version }}
+
+    steps:
+      - name: Install build dependencies
+        run: |
+          yum update -y
+          yum install -y epel-release centos-release-scl
+          yum install -y python3-pip ninja-build cmake3 pkgconfig gcc gcc-c++
+          python3 -m pip install meson
+          ln -s /usr/bin/cmake3 /usr/bin/cmake
+
+      - name: Download source tarball
+        uses: actions/download-artifact@v2
+        with:
+          name: source-tarball
+
+      - name: Extract source tarball
+        run: |
+          tar -xf "criterion-${version}.tar.xz" && rm "criterion-${version}.tar.xz"
+
+      - name: Build Criterion
+        run: |
+          meson setup -Dc_std=gnu11 --prefix /usr/local --libdir lib --buildtype plain \
+            --force-fallback-for libffi --wrap-mode nodownload \
+            build "criterion-${version}"
+          meson compile -C build
+          meson install -C build --skip-subprojects --destdir "$(realpath .)/binaries/criterion-${version}"
+
+      - name: Create binary tarball
+        run: |
+          tar --create --xz --file "binaries/criterion-${version}-linux-x86_64.tar.xz" \
+            --exclude='*.a' --transform="s,^\.,criterion-${version}," \
+            -C "binaries/criterion-${version}/usr/local" .
+
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: binaries
+          path: binaries/criterion-*.tar.xz
+          if-no-files-found: error
+
+  release:
+    name: "Release"
+    runs-on: ubuntu-latest
+    needs: [tarball, binaries]
+    env:
+      version: ${{ needs.tarball.outputs.version }}
+
+    steps:
+      - name: Check out Criterion
+        uses: actions/checkout@v2
+
+      - name: Download source tarball
+        uses: actions/download-artifact@v2
+        with:
+          name: source-tarball
+
+      - name: Download binaries
+        uses: actions/download-artifact@v2
+        with:
+          name: binaries
+
+      - name: Create pre-release
+        if: github.ref_type == 'branch'
+        run: gh release create --prerelease --target "${RELEASE_BRANCH}" --title "v${version}" "v${version}" criterion-*.tar.xz
+
+      - name: Create release draft
+        if: github.ref_type == 'tag'
+        run: gh release create --draft --target "${RELEASE_BRANCH}" --title "v${version}" "v${version}" criterion-*.tar.xz
--- a/meson.build
+++ b/meson.build
@@ -1,7 +1,7 @@
 project('criterion', 'c',
 		meson_version   : '>= 0.51.0',
 		license         : 'MIT',
-		version         : '2.3.3',
+		version         : '2.4.0',
 		default_options : ['c_std=c11', 'cpp_std=c++11', 'warning_level=2'])
 
 abi_version = '3.2.0'
--- a/meson.build
+++ b/meson.build
@@ -4,7 +4,7 @@ project('criterion', 'c',
 		version         : '2.3.3',
 		default_options : ['c_std=c11', 'cpp_std=c++11', 'warning_level=2'])
 
-abi_version = '3.1.0'
+abi_version = '3.2.0'
 
 # standard install directories
 prefix    = get_option('prefix')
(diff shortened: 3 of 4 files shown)


.github/workflows/version-bump.yml
```diff
@@ -0,0 +1,50 @@
...
+    steps:
+      - name: Check out Criterion
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Configure git
+        run: |
+          git config user.name "github-actions"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
Snaipe, line 32: Does this have any significance or should we just pick some Acme robot email?

.github/workflows/version-bump.yml
```diff
@@ -0,0 +1,50 @@
...
+
+      - name: Create branch for version bump
+        run: git switch -c "release/bump-v${VERSION}"
+
+      - name: Bump version
+        run: |
+          version_filter="(version\s*:\s*')[0-9a-z.+-]+'"
+          sed -E -i "s|${version_filter}|\1${VERSION}'|g" meson.build
+          sed -E -i "s|${version_filter}|\1${ABI_VERSION}'|g" src/meson.build
```
Snaipe, line 41: I don't think the ABI version should be updated alongside the version; for one, the ABI version is 3.x, not 2.x, and ABI changes are usually important enough for bumps to happen in the commits themselves.

.github/workflows/version-bump.yml
```diff
@@ -0,0 +1,50 @@
...
+          git config user.name "github-actions"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+
+      - name: Create branch for version bump
+        run: git switch -c "release/bump-v${VERSION}"
+
+      - name: Bump version
+        run: |
+          version_filter="(version\s*:\s*')[0-9a-z.+-]+'"
+          sed -E -i "s|${version_filter}|\1${VERSION}'|g" meson.build
```
Snaipe, line 40: I think I'd rather have this be fully inferred from git state. There's code in meson.build to run git describe to get the proper version for any given commit, so I think we should just put an obviously invalid fallback, and provide a way to freeze the version string for source tarballs without git metadata (e.g. reading the content of a .version file in the archive for instance). In particular, I would really like to avoid robot commits.

Snaipe (commented):
LGTM apart from the version bump; I think we can do better than having robot commits.

.github/workflows/version-bump.yml
MrAnno: That weird string seems to be the email address of @github-actions. I've added it so that commits from the bot appear correctly on the GitHub UI.

.github/workflows/version-bump.yml
MrAnno: I see. I'll remove that :)

.github/workflows/version-bump.yml
MrAnno: Okay, let's create a version file to freeze the version in the tarball.

Just to make sure I understand: do you prefer not to use automated git operations at all, or do you just want to avoid bot commits that do their own code modification?

For example, would that be acceptable for a bot to create an annotated tag, merge the master branch back to bleeding when the release is done, et cetera, or should we rather do all of those manually?

Snaipe: I'm not against having some git operations done by robots, but there aren't that many left if you remove commits and tags from the equation. Robot commits tend to pollute the history so I'd rather avoid them whenever possible, and I think tags shouldn't be done by robots either, because you can't do some things when tagging as a robot.

For once, you can't sign tags unless you breach trust and allow some external system access to your private key, or you have to generate a key pair for a robot, at which point you essentially tell others to trust a robot whose key management is not transparent rather than trusting a human maintainer. Though, now that I'm double checking, I actually forgot to sign the release candidate tag, but I'd rather have something that checks the validity of the tag rather than something that creates it for me, I think.

It also feels backwards to automatically create tags; what's the trigger for that operation? I'd have assumed that what would trigger all of the automation to build, package, and publish a release would have been a human pushing a tag.

Also, that's not to say that we can't have automation locally; I'm fine with a script to prepare a tag with the right message, perform any --no-ff merge, and whatever else is needed, since at the end of the day the operations would be authored by whoever runs the script.

MrAnno: > create a version file to freeze the version in the tarball.

It seems Meson won't allow uncommitted changes in the tarball, so we need to commit the version-freeze (at least as a local commit):
https://github.com/mesonbuild/meson/issues/4277

> It also feels backwards to automatically create tags; what's the trigger for that operation?

The trigger would have been a manual action, such as adding a label, or pushing a "release button".
My intention with this was to document the release process as code, which can be executed in CI.

I see your point, let's change the workflow:
- remove the release bump workflow
- figure out what the "release draft workflow" trigger should be

I think a tag-push trigger would be too late for us because the source tarball or binary generation may fail due to a real bug in Criterion, which then we want to fix before the release. I believe tags should not be removed/re-created, so I think the best would be if we triggered the workflow with something else, for example:

trigger by
- pushing a commit to the master branch with a specific instruction in it, such as "@github-actions draft release v2.4.0"
- sending an instruction as a comment in a GitHub issue
- always creating a pull request manually with a "release" label and merging it
- starting the job manually on GitHub, where you can input the version number (since we don't have a tag yet)

Snaipe: I also agree that tags should not be removed or recreated, but given that we have a release candidate to fix any issues before an actual release, I think we should make the most out of it. How about this:

* Every time new commits gets pushed to master, the new branch tip gets a release (as in, published on the github release page), alongside any built packages. That release is always marked as a prerelease, and its version is generated from `git describe --long`.
* If a new tag not containing `-rc` is pushed, a new release (which this time isn't marked as a prerelease) gets published.

If we do this, then it means that the release workflow becomes:

1. merge bleeding into master, tag it with `vX.Y.Z-rc`, push both master and the tag.
2. the release action runs, builds a release source tarball & binary build for Criterion on CentOS, then creates a new prerelease on github based of the output of `git describe --long`, which will look like `vX.Y.Z-rc-0-gdeadbeef`, and attached the aforementioned artifacts to that release.
3. someone makes a pull request to fix a bug in the vX.Y.Z release candidate, we review & merge it to master; let's say for this argument that it contains 3 new commits and the last one is `feedface`
4. the release action runs because master has new commits, and goes on to make a new prerelease named `vX.Y.Z-rc-3-gfeedface`.
5. after some time, when the prerelease feels ready for release, we tag vX.Y.Z on the tip of master, which causes an action to run and create the final release named vX.Y.Z.

Some additional would-be-nice-to-have on top of that:
* automatically fast-forward bleeding to be on the tip of master. If that's not possible, warn the maintainers to manually intervene if necessary (not being able to fast-forward isn't necessarily bad, it could be that development continued on bleeding and we cherry-picked a few commits from bleeding to master)
* check that the release tag is signed; warn maintainers if not (don't think we can do much more than warn given that changing tags is not allowed, although we could make an exception for the automation and automatically remove it precisely because we deem unsigned release tags to be invalid)

What do you think?

MrAnno: Sounds good to me. I'll try all this in my fork in practice and share the results.

MrAnno: My plan is to try all the above over the weekend.

MrAnno: I've updated the description according to my changes. I'll test the workflow properly and attach a few screenshots tomorrow.

MrAnno: I've updated the description according to my changes. Tomorrow, I'll test the workflow properly and attach a few screenshots.

Meson supports version files, but that file must always be present and the feature is only available in new Meson versions, so I've decided to leave the version string intact. A manual version bump is needed before each release, but this is not a hard requirement, because the tarball is created using `meson rewrite kwargs set project / version`, which replaces the version number in `meson.build`.

MrAnno:
Done. I've attached new screenshots too.

MrAnno:
@Snaipe Let me know if you want to try it in my fork.

Snaipe:
Will review this now; I'd be interested in seeing what happens (more specifically the resulting entries in the release page) when you tag something like `v1.2.3-rc` in your fork, then make a new commit on top of master, then tag that commit as `v1.2.3`.

.github/workflows/ff-bleeding.yml
```diff
@@ -0,0 +1,39 @@
...
+    steps:
+      - name: Check out Criterion
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Configure git
+        run: |
+          git config user.name "github-actions"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
Snaipe, line 26: This is fine, but I'm curious where it's actually used now that we've removed any git object-writing operation; does the github actions robot user get used by `gh issue create` maybe?

.github/workflows/release.yml
```diff
@@ -0,0 +1,158 @@
...
+  push:
+    branches:
+      - master
+    tags:
+      - v[0-9]+.[0-9]+.[0-9]+
+
+env:
+  RELEASE_BRANCH: master
+  RELEASE_CANDIDATE_TAG_PATTERN: ^v[0-9]+.[0-9]+.[0-9]+-rc
+  RELEASE_SIGNING_KEY_ID: 75B0 21D9 64AD 8693 24CD  B74D DDD7 BCCD 44BB 9A9D
```
Snaipe, line 14: Is this your key? Might be worth adding the ability to receive multiple keys and verifying that one of the keys signed the commit, that way you and I can individually sign release tags if needed without sharing some common key.

.github/workflows/release.yml
```diff
@@ -0,0 +1,158 @@
...
+          fi
+          version="${version:1}"
+          echo "${version}"
+          echo "::set-output name=version::${version}"
+
+      - name: Create source tarball
+        env:
+          version: ${{ steps.detect-version.outputs.version }}
+        run: |
+          meson rewrite kwargs set project / version "${version}"
```
Snaipe, line 76: This is really nice!

.github/workflows/ff-bleeding.yml
```diff
@@ -0,0 +1,39 @@
...
+  push:
+    branches:
+      - master
+
+env:
+  FF_ISSUE_TITLE: "No-FF: bleeding is behind master"
+  GITHUB_TOKEN: ${{ github.token }}
+
+jobs:
+  release-tag-verification:
```
Snaipe, line 14: Should probably be named fast-forward-bleeding

Snaipe (approved):
This is great, thanks for doing this. I have a couple of nits but this overall looks good to me.

MrAnno:
Yep, that is correct. In order to start producing release candidates, one has to push an annotated tag "vX.Y.Z-rc" first.
Master pushes will stop generating release candidates once the real release tag (vX.Y.Z) is published.

I've prepared my Criterion fork to test this. It is now in release candidate mode, you can try the process by pushing commits to my master branch (https://github.com/MrAnno/Criterion/). If you're okay with signing a fake release tag with your keys there, we can try the tag validation job as well.

If you add something to `bleeding` and then push something onto `master`, the "fast-forward bleeding" job will open an issue as a reminder.

.github/workflows/ff-bleeding.yml
MrAnno: Oops, indeed. Thank you.

.github/workflows/release.yml
MrAnno: This is your key. We can add mine too later if that's okay with you.

.github/workflows/ff-bleeding.yml
MrAnno: I think `gh issue create` works without this, I left it here because we still have to create an unpublished local commit, because Meson refuses to create a source tarball from uncommitted changes.

MrAnno:
@Snaipe I've added some more examples to my fork:
https://github.com/MrAnno/Criterion/releases

I've set up a scenario where bleeding could not be fast-forwarded, here is the result:
https://github.com/MrAnno/Criterion/issues

My fork is still in "release candidate" mode.  Would you like to push a signed annotated tag (`v0.0.0`) to test the release tag signature validation?

Snaipe:
I'll try testing it later today -- Embarrassingly enough, I can't find my yubikey, so I need to dig out my GPG key from my previous laptop.

MrAnno:
Thank you.

The same happened to me a couple of months ago. :)

Snaipe:
I finally got my keys back in order, and I'm currently testing the workflow.

One minor thing I've noticed is the binary tarball has some oddballs within:

```
criterion-3.4.5-rc-0-g80ea92c/usr/lib64/libcm_nanomsg.a
criterion-3.4.5-rc-0-g80ea92c/usr/lib64/libcm_protobuf_nanopb_static.a
criterion-3.4.5-rc-0-g80ea92c/usr/lib/libcriterion.a
criterion-3.4.5-rc-0-g80ea92c/usr/lib/libcm_git2.a
```

I'm debating whether we want to release a static version of libcriterion, given that it's also missing BoxFort, and we'd have to ship a bundle of static libs to make it work. I'm not sure it works, but let me test it and get back to you.

If we want, in fact, to release a static version, I think we need to get the libdir prefix in order, it seems wrong that some would go in lib64 while the others go in lib.

Another thing I'm wondering if whether the /usr prefix makes sense here. We certainly don't want people to be stomping on /usr, and extracting the release somewhere only adds another layer. I think it'd make more sense if we got rid of the usr component and simply had `criterion-3.4.5-rc-0-g80ea92c/lib/libcriterion.so`, `criterion-3.4.5-rc-0-g80ea92c/include/...`, etc.

Snaipe:
Using the static lib seemed to have worked, although the ldflags are pretty gnarly:

`gcc ./samples/simple.c -Icriterion-3.4.5-rc-0-g80ea92c/usr/include -Lcriterion-3.4.5-rc-0-g80ea92c/usr/lib64 -Lcriterion-3.4.5-rc-0-g80ea92c/usr/lib -lffi -lcriterion -lanl -pthread -lrt -lm -lpthread -lcm_git2 -L/usr/lib64 -lcm_nanomsg -lcm_protobuf_nanopb_static`

Thankfully these were mostly generated by pkg-config so it's technically not that bad, but the installation of vendored libgit2, nanomsg, and nanopb seemd weird. It's even weirder considering that libcritrion.a apparently contains boxfort but not these extra libs.

Snaipe:
Aside from all that, I've just finished a full workflow and it properly created the releases and everything, so I think it looks good to go.

MrAnno:
Yay! Thanks for trying out the workflow.

I'll try to address the remaining problems soon.

About the `/usr` prefix: Previous Criterion binary packages had an `/usr` prefix in their pkg-config file, which could have been problematic when those binaries are installed anywhere else.

We could go with a simple `/` prefix, which will produce the desired directory structure, but the pkg-config prefix will then be `/` too, which might be weird. Or alternatively, we can use the default prefix (`/usr/local`), and just skip `/usr/local` in the `tar` command.

Snaipe:
The issue is that the published binary release is technically relocatable (or rather, isn't relocatable right now because of pkg-config but should be). It shouldn't really matter whether the user extracts it under /usr/local or ~/.local, I think it should work the same in both ways.

pkg-config files can be made [relocatable](https://gitlab.freedesktop.org/pkg-config/pkg-config/-/blob/master/pkg-config.1#L400) by specifying the prefix to be relative to the .pc file. I wonder if we should be doing this via some post-processing.

With that being said, I'm more and more skeptical by the idea of manually providing binary releases outside of package management systems, so I think we can just fix the prefix to be /usr/local and move on. Leaving /usr is encouraging users to extract the release into /usr, and stomp all over the domain of their local package manager.

MrAnno:
The static lib issue seems to be a bit complicated:
- I could bundle all CMake-based static subproject dependencies into `libcriterion.a`, if we bumped the min. Meson version to 0.55.0. (See: [CMake `set_install()`](https://mesonbuild.com/CMake-module.html)), but
- currently, there is no method for native Meson subprojects (libffi) to skip installation (and to bundle static deps).

I think the easiest would not be to provide static libraries for now in the binary package.

----

It turned out that Criterion is currently linked dynamically to the system's `libffi` in the binary package, which needs to be fixed.
I'm trying to fix the compilation of the libffi subproject so that we can link to it statically.

MrAnno:
Done. The binary package looks much better now:
https://github.com/MrAnno/Criterion/releases/tag/v0.0.0-rc-0-g4cccb18

Snaipe:
LGTM. Feel free to merge when you think this is ready.

MrAnno:
Okay, I think it's ready.

Thanks for the persistent and thorough review and for the ideas/advice.

PR #416, opened 2022-01-11
build: add -Drpath-in-pkgconfig=enabled build option 

### LINKED ISSUE #415: rpath needed in criterion.pc (v2.4-rc)
Hello,

criterion.pc does not add `-Wl,-rpath=${libdir}` to `Libs:`. When I install criterion to a non-default location then I can link but not execute my tests, because libcriterion.so is not found at runtime.

I can't remember that this was a problem in 2.3.3 but I didn't double check that yet.

MrAnno:
This is technically a feature, not a bug fix for the upcoming release, but I've opened the PR against "master" in the hope that this could be the last feature added for v2.4.0.

Resolves #415


Snaipe (approved):

kugel-:
Thanks for the effort. I'm not entirely happy though with requiring the user to explicitly enable that option (in addition to him setting -Dprefix).

In my view, /lib and /usr/lib are the standard libdirs, and none else, and these do not require rpath (and is discouraged and sometimes even enforced by distros). For all others rpath could be automatically set.

(One could argue that /usr/local/lib also doesn't need rpath as this dir can be assumed to be administratively added to the dynamic linker search path, I don't feel strong about that one)

Snaipe:
> In my view, /lib and /usr/lib are the standard libdirs, and none else, and these do not require rpath (and is discouraged and sometimes even enforced by distros). For all others rpath could be automatically set.

Maybe the default for the option could be true (i.e. define it as `type: 'feature', value: 'auto'` & do runtime checks on the libdir) iff the libdir isn't /lib or /usr/lib. I don't feel too strongly about the issue, but I'd rather have a default that works for the majority of use-cases.

MrAnno:
I think a few other paths can be listed as well as "system path" if we consider Homebrew on macOS, or NixOS. That is the main reason I would avoid maintaining a list of what we consider as "system paths".

kugel-:
Maybe Criterion is in a special position. Running automated unit tests (aka uninstalled programs) doesn't work by default, without RPATH. Things aren't that bad for installed programs (libtool usually sets rpath in the executable if the programs are installed into the same non-standard prefix). LD_LIBRARY_PATH could work but is also something you'd set as a user when "interactively" running the program but not so much for automated tests during `make check`. One would have to go great length to fixup test runners to consider LD_LIBRARY_PATH and set it to the right value.

Snaipe:
> I think a few other paths can be listed as well as "system path" if we consider Homebrew on macOS, or NixOS. That is the main reason I would avoid maintaining a list of what we consider as "system paths".

Yeah, good point. Unless we have some reliable way to know the list of system paths to figure out whether to use rpath or not, this is going to be hard to do right. Even if we figure out such a list at build time, it's also possible that the actual list might have changed by the time criterion gets linked somewhere (especially with nix).

I mean, ultimately the rpath problem is a packaging problem. I would argue that we should strive to not have to set rpath by having criterion packaged on all platforms our users care about, and if you want to install it anywhere else, you have to know what you're doing, including using -Dset-rpath=true. In that model I think I'm happy with what we have here.

kugel-:
I can understand that reasoning. Maybe special-case at least /home/xxx-prefixes, then? Clearly those are per-user installation that are always going to need rpath. For others explicit `-Dset-rpath=true` could be required.

Snaipe:
making the default for set-rpath true when the libdir starts with /home seems like a reasonable compromise to me. @MrAnno thoughts?

kugel-:
FWIW, the name of the option ("set-rpath") could be a bit missleading, thinking it would control the setting of RPATH in libcriterion.so. Maybe "rpath-in-pkg" or something like that is clearer?

MrAnno:
Adding a `/home` check sounds a bit unorthodox to me, but I'll do that happily if you @kugel- finds it useful.
If someone else asks for another exception (`/tmp`, for example), let's just say we stop at 3 :)

@kugel- I'll update the pull request with the `/home` check.

Just out of curiosity, could you share your exact use case? (For example, the syslog-ng development team also uses Criterion internally in their own CI, but Criterion is usually preinstalled into a Docker image, where ld.so.conf is configured to `/usr/local/lib` and `/opt/secret-ci-stuff/lib`.)

> Maybe "rpath-in-pkg" or something like that is clearer?

I like the name, I'm going to rename the option.

kugel-:
Well when I opened #415 I installed it locally under /home/kugel/... Well prefix below /home is about the only real "per-user" installation possiblity on unix-based systems, right? 

My use case:
I wanted to see if Criterion 2.4 breaks any of our existing tests. I have installed criterion v2.3.3 system-wide (Debian packages provided by @Snaipe )  so I didn't not want to interfere with that. I usually install packages from source right into the build directory for quick tests (e.g. `-Dprefix=$PWD/build/dest`) and then refer to that via pkg-config, e.g. I build the tests using `./configure PKG_CONFIG_PATH=/home/kugel/.../dest/lib/pkgconfig && make && make check`

I also tend to make other per-user installation to $HOME/.local

MrAnno:
I see, thank you for the explanation.

MrAnno:
I've added the `/home/` check and renamed the option to `-Drpath-in-pkgconfig=enabled`.

MrAnno:
@Snaipe What do you think, can this go in? The name has slightly changed: -Drpath-in-pkgconfig=enabled/disabled, default is  "auto" with /home detection.

src/meson.build
```diff
@@ -104,10 +104,15 @@ criterion = declare_dependency(
 	dependencies: deps,
 	link_with: [libcriterion])
 
+pkgconfig_rpath = ''
+if get_option('rpath-in-pkgconfig').enabled() or (get_option('rpath-in-pkgconfig').auto() and  prefix.startswith('/home/'))
```
Snaipe, line 108: micro-nit: extra space in `and  prefix.startswith('/home/')` after the `and`

Snaipe (approved):
LGTM, I just have one tiny nit, but feel free to merge this yourself once you've fixed it.

src/meson.build
MrAnno: Fixed. Thanks for noticing.

PR #422, opened 2022-01-31
fix small grammatical error

japrozs:


Snaipe:
Nice catch! Thanks for the fix.

We're currently in the release candidate cycles, so to avoid any merge shenanigans between bleeding and master, I'll manually cherry pick this on top of master.

Snaipe:
Rebased and merged as 27ab2ca2c310de31a309aa70b13bcdfcefd9b7b6.

japrozs:
> Nice catch! Thanks for the fix.
> 
> We're currently in the release candidate cycles, so to avoid any merge shenanigans between bleeding and master, I'll manually cherry pick this on top of master.

gotcha

Issue #364, opened 2020-11-11
Criterion does not work wsl ubuntu


Nekyt:
Hello!
I need to use criterion for my school project.
I successfully built it from source and installed it using meson &ninja and pulled from the latest repo with the -recursive option.
Unfortunately when i execute criterion I see nothing in my shell on wsl ubuntu. 
The fan starts working very loud and that's it.
I even compiled a version of criterion natively from a valid linux distro on which criterion worked well and tried to execute the ./test
on wsl but it still doesn't show anything wether it is in gnome shell or simple windows bash.
this is a catastrophe!
Please help

Nekyt:
i ve got other pb now

Nekyt:
now i have 
mincore(2) returned an unexpected error: Function not implemented  This is a bug; please report it on the repository's issue tracker.         Aborted (core dumped)
thanks in advance

Snaipe:
Do you still have the original issue?

Mincore returning -ENOTSUP is a bit infuriating -- why would WSL export the function if it's not going to be implemented?

Anyway, can you try to rebuild boxfort with `#mesondefine HAVE_MINCORE` removed in config.h.in? That should force boxfort to use msync instead.

Nekyt:
I am sorry but the command 
 grep -inr "HAVE_MINCORE" .  
returns  no result at all.
I don't understand what to remove.

Nekyt:
proffessor Snaipe! 
thank you very much for your help.
after pulling from master i was able to successfully install criterion from source on wsl.
I would like to point out that I was unable to install criterion from ppas using apt.
People at my school use a script which install criterion on Fedora 33 automatically, which I unfortunately couldn't use because...
unless you use this script and you are not a pro in c &linux it is hard to install.
The release is broken and I had to switch branches on git.
I understand that you have family kids etc and that the project is free of charge, but it is good that you as pros know that down on earth it is hard to install. Thank you anyways
Your code is amazingly clean: no memory leaks!!(never happened to me)

r3dapple:
Im running in the same problems Nekyt ran into. I installed from source according to this: https://criterion.readthedocs.io/en/latest/setup.html#installation fixed the issue of not finding the dynamic lib with this:
`g++ test.cpp -lcriterion -Wl,-rpath,/usr/local/lib/x86_64-linux-gnu`

[...]

PR #425, opened 2022-02-19
Workaround for unimplemented mincore() on WSL 1

### LINKED ISSUE #424: wsl,boxfort: mincore fails with ENOTSUP in arena code
A couple of users on WSL seem to be hitting the following error (#364)

```
mincore(2) returned an unexpected error: Function not implemented This is a bug; please report it on the repository's issue tracker. Aborted (core dumped)
```

It's fairly disappointing to see that the WSL glibc goes out of its way to export mincore, but ends up not implementing it. That said, if we assume that our debian packages will end up running on WSL at some point, it might be reasonable enough to fix this in BoxFort.

I don't know if there's a separate WSL branch of ubunu but if there isn't, having a feature toggle to disable the mincore check on WSL might not be sufficient, and instead msync should always be used as fallback to mincore on ENOTSUP.

MrAnno:
Fixes #424
See also #364

Depends on Snaipe/BoxFort#38

Note: I've added `[skip actions]` to the commit message just in case we want to skip creating another release candidate.

IMHO no changelog entry is needed, `Fix: worked around msync not properly working on WSL and causing the runner to break` covers this PR as well.

### DIFF
--- a/subprojects/boxfort.wrap
+++ b/subprojects/boxfort.wrap
@@ -1,4 +1,4 @@
 [wrap-git]
 directory = boxfort
 url = https://github.com/Snaipe/BoxFort.git
-revision = fe06439edd619ef574cf897195f0976bb60e4781
+revision = a26074c5cc900fabc85424253f0f64e3b9c661c3


MrAnno:
Test suite on WSL 1:

Before
```
45/45 criterion / full / theories_regression.cc  FAIL     0.09 s (killed by signal 6 SIGABRT)

Ok:                    3
Expected Fail:         0
Fail:                 42
Unexpected Pass:       0
Skipped:               0
Timeout:               0
```

After
```
45/45 criterion / full / theories_regression.cc  OK       0.15 s

Ok:                   45
Expected Fail:         0
Fail:                  0
Unexpected Pass:       0
Skipped:               0
Timeout:               0
```

Snaipe:
LGTM. The fix is trivial, so let's not make this delay the v2.4.0 release -- once this gets merged I'll tag the new version.

PR #430, opened 2022-02-24
diff: fix leak due to not calling git_libgit2_shutdown

### LINKED ISSUE #429: cri_diff_init leaks memory
According to [Valgrind](https://valgrind.org/), [`cri_diff_init`](https://github.com/Snaipe/Criterion/blob/4cf886a2d96ada425e8c88d3c4d935e510c7c712/src/string/diff.c#L38) leaks memory, e.g.:

```
==123456== 96 bytes in 1 blocks are still reachable in loss record 18 of 18
==123456== at 0x4837B65: calloc (vg_replace_malloc.c:752)
==123456== at 0x48DC368: git__calloc (util.h:169)
...
==123456== at 0x486CB3B: git_libgit2_init (global.c:278)
==123456== at 0x486405F: cri_diff_init (in libcriterion.so.3)
```

`cri_diff_init` calls `git_libgit2_init`, which should be accompanied by a call to [`git_libgit2_shutdown`](https://libgit2.org/libgit2/#HEAD/group/libgit2/git_libgit2_shutdown) when finished with the `libgit2` context. I could not find any calls to `git_libgit2_shutdown` in the code, so that would explain the leaks.

#### Appendix

Test program `test_foo.c`:
```c

Test(foo, bar) {
    cr_assert(0);
}
```

Compiled with `gcc (Debian 8.3.0-6) 8.3.0)` and linked against [Criterion 2.4.0](https://github.com/Snaipe/Criterion/releases/tag/v2.4.0) on `Debian 4.19.194-3 (2021-07-18) x86_64`:
```
gcc -g -O0 test_foo.c -o test_foo -lcriterion
```

Ran test program `test_foo` under Valgrind:
```
valgrind -v --track-origins=yes --leak-check=full --show-leak-kinds=all ./test_foo
```

Snaipe:
Fixes #429.

### DIFF
--- a/src/core/runner.c
+++ b/src/core/runner.c
@@ -242,6 +242,7 @@ CR_API void criterion_finalize(struct criterion_test_set *set)
     VALGRIND_ENABLE_ERROR_REPORTING;
 #endif
 
+    cri_diff_fini();
     criterion_free_output();
 }
 
--- a/src/core/runner_coroutine.c
+++ b/src/core/runner_coroutine.c
@@ -234,6 +234,7 @@ static int run_test_child(void)
 #endif
 
     cri_proto_close(g_client_socket);
+    cri_diff_fini();
 
 #ifndef ENABLE_VALGRIND_ERRORS
     VALGRIND_ENABLE_ERROR_REPORTING;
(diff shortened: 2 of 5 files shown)


g1g4b0t:
Looks good to me.

MrAnno (approved):

PR #440, opened 2022-04-05
theories: reset ffi argument vector on each iteration

### LINKED ISSUE #437: Theories not working
Hey!

I tried to add theories in a project of mine but the permutations don't seem to work.

As you can see on the screenshot below, the values are always the same (the first two, both 0s).

in
out

And I am following the doc.

Am I missing something?

Snaipe:
This regression was introduced by the recent move to libffi.

Fixes #437.

### DIFF
--- a/samples/theories.c
+++ b/samples/theories.c
@@ -26,6 +26,7 @@ TheoryDataPoints(algebra, bad_divide_is_inverse_of_multiply) = {
 
 Theory((int a, int b), algebra, bad_divide_is_inverse_of_multiply) {
     cr_assume(b != 0);
+    cr_assume(a != INT_MIN || b != -1);
     cr_assert_eq(a, bad_div(bad_mul(a, b), b));
 }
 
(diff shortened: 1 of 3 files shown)


MrAnno:
LGTM. Something is not right with the Windows tests, I'll look into it.

Snaipe:
I've had some time to investigate something that I had a hunch on -- I was noticing that on linux I was getting a SIGFPE on the test case involving dividing INT_MIN with -1. This is UB, and while catching the crash on linux is great and proving that this is helping users catch these issues as intended, I also stumbled on [this reddit post](https://www.reddit.com/r/learnprogramming/comments/414x8k/c_what_happens_when_i_divide_int_min_from_the/) whose author reports their program freezing.

Given that most people use Windows, it's entirely possible that for some reason MSVC++ is emitting a set of instructions that causes the program to get stuck when performing the division. That, or the signal handling ends up being slow enough that the test times out.

Snaipe:
Anyway, the fix should be straightforward enough, so I'll merge it as is and make a release candidate for v2.4.1.

PR #452, opened 2022-06-19
Fix user-facing compiler warnings with `-Wconversion`

### LINKED ISSUE #446: Compiler Warnings using -Wconversion
### Discussed in https://github.com/Snaipe/Criterion/discussions/441

<div type='discussions-op-text'>

<sup>Originally posted by **senarclens** April  6, 2022</sup>
After evaluating multiple unit test libraries for C my conclusion was that criterion might be the best available choice. However, I'm a bit unsure why there needs to be a compiler warning when using `gcc -Wconversion` as soon as assertions are used. For example using `cr_assert` triggers

```
warning: conversion from ‘int’ to ‘unsigned char:1’
```

regardless of the content of the assertion. Looking at the source I found that the warning is due to the assert node having a 1 bit bitfield for its `pass` member.

```
struct cri_assert_node {
    const char *repr;
    const char *message;
    struct cri_assert_param params[CRI_ASSERT_MAX_PARAMS];

    unsigned pass : 1;
    unsigned dynrepr : 1;
    uint32_t nchild : 30;
    uint32_t maxchild;
    struct cri_assert_node *children;
};
```
`pass` gets assigned a regular int (`cri_tmpn.pass = cri_cond_un;` after the preprocessor is done) leading to the warning.
I'm not aware of any way of preventing the warning other than increasing the size of pass to, eg, int (or char) or using a single integer for `pass`, `dynrepr` and `nchild` and applying bitmasks when reading or assigning them. I'm aware that the former would be a waste of 32 (or 8) additional bits of memory, the latter would require working with bitmasks. At the same time, I am far from sure whether it is worth having a warning for that amount of memory (or using bitmasks instead of bitfields) when using gcc (version 9.4.0 on my system) with the `-Wconversion` flag. I always use `-Wconversion` for my code to avoid unintentional mistakes and it would be lovely to keep it that way for my unit testing code as well.
So, out of honest curiosity, can someone please tell me the reasoning behind using bitfields in cri_assert_node (eg. the saved bits having a really important impact on embedded systems or the like + why bitmasks would not be an alternative either)?
Is there any way of getting rid of that message without disabling `-Wconversion` in gcc?
Thanks a lot!

</div>

MrAnno:
Fixes #446.

There are ~160 warnings in Criterion related to implicit conversions.
I haven't fixed all of them as most seemed to be false positives.
This PR focuses on user-facing ones and a few others.

### DIFF
--- a/include/criterion/internal/assert/op.h
+++ b/include/criterion/internal/assert/op.h
@@ -316,7 +316,7 @@
                 ) ")";                                                      \
             size_t cri_paramidx = 0;                                        \
             CRITERION_APPLY(CRI_ASSERT_IT_MKNODE_AUTO, , __VA_ARGS__)       \
-            cri_tmpn.pass = cri_cond_un;                                    \
+            cri_tmpn.pass = !!cri_cond_un;                                  \
             cri_prevnode = cri_assert_node_add(cri_node, &cri_tmpn);        \
         }                                                                   \
     } while (0)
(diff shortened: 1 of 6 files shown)


Snaipe (approved):
Thanks for fixing this -- LGTM.

PR #464, opened 2022-10-25
diff: fix crash in assert failure reporting when there is no diff

### LINKED ISSUE #463: Printing new lines in cr_user_*_tostr function results in SIGSEGV
```c
#include <criterion/new/assert.h>

struct dummy_struct {
    char a;
    size_t b;
};

/* We need to provide basic functions for our dummy struct */
int cr_user_dummy_struct_eq(struct dummy_struct *a, struct dummy_struct *b)
{
    return a->a == b->a && a->b == b->b;
}

char *cr_user_dummy_struct_tostr(struct dummy_struct *d)
{
    char *out;

    cr_asprintf(&out, "hello\n");
    return out;
}

Test(asserts, array) {
    struct dummy_struct s1 = {4, 2 };
    struct dummy_struct s2 = {4, 7 };

    cr_assert(eq(type(struct dummy_struct), s1, s2));
}

```
```sh
$ gcc main.c -l criterion && ./a.out
[----] main.c:27: Assertion Failed
fish: Job 1, './a.out' terminated by signal SIGSEGV (Address boundary error)
```

Snaipe:
If the user-provided function to stringify an object is bugged and produces equal output for non-equal objects, thus producing an empty diff, Criterion would crash trying to access an unset parameter list.

This fixes the crash by checking that we do have expected values, and printing placeholder messages if we do not.

Fixes #463.

### DIFF
--- a/src/core/client.c
+++ b/src/core/client.c
@@ -543,13 +543,16 @@ bool handle_assert(struct server_ctx *sctx, struct client_ctx *ctx, const criter
     if (asrt->results_count > 0) {
         for (size_t i = 0; i < asrt->results_count; ++i) {
             criterion_protocol_result *res = &asrt->results[i];
-            if (res->value.params->list_count == 0 && !res->message) {
-                continue;
-            }
 
             log(assert_sub, &asrt_stats, res->repr, res->message);
 
-            if (res->which_value == criterion_protocol_result_params_tag) {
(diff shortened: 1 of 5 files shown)


MrAnno (approved):
Sorry for the delay.

MrAnno:
Happy new year :)

PR #468, opened 2023-01-02
Build and CI-related chores

MrAnno:
Depends on Snaipe/BoxFort#45


MrAnno:
I'm working on debugging the "timeout" test failure.
I believe Cirrus CI caught a very rare and long-standing Criterion bug that I saw time-to-time in my dev environment too.

MrAnno:
> I believe Cirrus CI caught a very rare and long-standing Criterion bug, which I saw time-to-time in my dev environment too.

It seems Criterion hangs in an `nn_recv()` call indefinitely because the responder process is sometimes killed too early.
I've pushed a "safety precaution" patch to avoid indefinite hanging in the future, and I'm working on fixing this particular issue properly.

MrAnno:
Yeah, it is a deadlock situation:

```
thread 0
--------
run_tests_async():
  while (read_message()):
    process_message()
    if not alive -> remove_client() -> destroy_client_context() -> bxf_wait() -> pthread_cond_wait(&instance->cond)

thread 1
--------
child_pump_fn():
  reap_child() -> death_callback() -> cr_send_to_runner() * 2 -> nn_recv()
  ...
  pthread_cond_broadcast(&instance->cond);
```

The main message loop waits for the sandbox to "complete", but the sandbox is considered "complete" only when its death callback is finished. Unfortunately, the death callback contains 2 `cr_send_to_runner()` invocations, which wait for the main message loop to send an ack, which will never happen -> deadlock.

MrAnno:
I think the root cause is the way how the `alive` handling works in Criterion (BoxFort behaves correctly in my opinion):

To indicate that a client is dead, a `death` message has to be sent to the runner, and that message needs to be the very last message. This is done correctly.

The problem is `handle_birth()`, which sets the `alive` flag to true. Under rare circumstances (overloaded system), the `handle_birth()` action may never run due to the test timeout that occurs before sending out the birth message. This scenario leaves the `alive` flag on false, causing the deadlock.

MrAnno:
I'll fix the deadlock in a separate PR.

PR #472, opened 2023-01-19
section: fix segfault if map_shdr fails

dgloe-hpe:
The get_section_data function doesn't check the return value of map_shdr, causing it to segfault in the following strcmp. Add a check so we return 0 rather than segfaulting.


MrAnno (approved):
Thank you.

PR #474, opened 2023-02-05
nanopb: fix incorrect arguments to protoc

bazsi:
In gen-pb.py, we call protoc to compile our protobuf declarations with nanopb's generator plugin.  The issue is that we used a space between -f and the name of the option file, which causes the option file to not be found as the space is considered part of the filename.

If the option file is not found, all our unit tests crash on Debian testing, with nanopb 0.4.7.  Earlier versions of nanopb probably were not so picky about that space.

Signed-off-by: Balazs Scheidler <bazsi77@gmail.com>


MrAnno (approved):
Wow :)

MrAnno:
All CI failures are related to #468.

PR #483, opened 2023-03-21
core: fix deadlock in overloaded scenarios

MrAnno:
Root cause: `client_ctx::alive` handling and BoxFort timeout conflict

To indicate that a client is dead, a death message has to be sent to the runner, and that message needs to be the very last message. This is done correctly.

The problem is `handle_birth()`, which sets the `alive` flag to true. Under rare circumstances (overloaded system), the `handle_birth()` action may never run due to the test timeout that occurs before sending out the birth message. This scenario leaves the `alive` flag on false, causing a deadlock:

```
-- thread 1
run_tests_async():
  while (read_message()):
    process_message()
    if not alive -> remove_client() -> destroy_client_context() -> bxf_wait() -> pthread_cond_wait()

-- thread 2
child_pump_fn():
  reap_child() -> death_callback() -> cr_send_to_runner() * 2 -> nn_recv()
  ...
  pthread_cond_broadcast(&instance->cond);
```

The death callback contains a `cr_send_to_runner()` invocation, which waits for the main message loop to send an ack, which will never happen.

### DIFF
--- a/src/core/client.c
+++ b/src/core/client.c
@@ -263,7 +263,7 @@ bool handle_birth(struct server_ctx *sctx, struct client_ctx *ctx, const criteri
     (void) sctx;
     (void) msg;
 
-    ctx->alive = true;
+    ctx->dead = false;
     return false;
 }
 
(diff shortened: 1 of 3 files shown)


MrAnno:
@Snaipe I think with this PR and #474, we should prepare a v2.4.2.

PR #487, opened 2023-03-30
Fix gt() and ge()

### LINKED ISSUE #473: gt assertion misbehaves in new assertion API
### Environment
Arch Linux with Criterion v2.4.1 (from [AUR](https://aur.archlinux.org/packages/criterion)).

### Issue
The following assertion passes, which is not the expected behavior.
```
#include <criterion/new/assert.h>

Test(base, test) {
    int x = 0;
    cr_assert(gt(int,x,0));
}
```

### Follow-up
Looking at the generated code (`gcc -E`), I found the following:
```
          cri_cond_un =
            ((!(cr_user_int_lt (&(cri_actual), &(cri_reference)))));
```
which is not correct, as `!<` is equivalent to `>=`, not to `>`.

Similar code is generated when using other tags, so this is not exclusive to the `int` tag.

Without any tags the condition becomes:
```
cri_cond_un = cri_actual > cri_reference;
```
which is correct.

MrAnno:
Before this patch, the following assert was true:
`cr_assert(gt(int,0,0));`

Fixes #473


Snaipe (approved):
Thanks, LGTM. Can you also add the same test case for asserts.cc? I don't think it was broken in C++, but it adds some coverage for the bug.

MrAnno:
Good idea. I've added it. Thank you for the review.

Issue #488, opened 2023-03-30
Include debugbreak and klib in source zip file


cprieto:
Hi!

I had been working in a recipe to include this amazing library in the Conan Index so it can be used with Conan (https://conan.io/), so far it is a work in progress... But I had been facing some problems regarding including the submodule dependencies (debugbreak and klib), mostly because the policies and checks in the public repository for recipes.

I know this could be a too much to ask thing, but could we include the source for those given revisions of those submodules in the zip source file provided with each release? at least for future versions?

Keep the great work!

MrAnno:
I'll take a look at this. I think we should just convert those 2 submodules to Meson subprojects.

Thanks for working on the Conan recipe.

cprieto:
Oh thanks @MrAnno I thought about doing exactly the same with a patch for the main `meson.build` file and include wraps around it. I had not finished my patch yet but if so, I will submit the PR soon, would that be ok for you?

Would this be ready for the upcoming 2.4.2?

cprieto:
PR submitted, only a problem with the required version for Meson in some build servers (Meson needs to be at least 0.55, we can include that restriction in the `project` definition). Besides that, it works without the need of submodules. :smile:

PR #489, opened 2023-03-31
Treat debugbreak and klib as external subprojects

cprieto:
Dealing with Git submodules is a little cumbersome for newcomers trying to compile the source code, at the same time, it provides challenges when integrating with some C++ dependency systems (like Conan).

Instead of depending on git submodules, this PR moves `debugbreak` and `klib` as external wraps from the corresponding GitHub revisions and repositories, the source code is unmodified and it uses meson and wrap support for overrides to create a dependency in `klib` and `debugbreak` include directories.

This will close https://github.com/Snaipe/Criterion/issues/488


cprieto:
The reason both Alpine builds (and one Debian build) are failing is because the version for Meson is too old. Notice it passes in Alpine for ARM and Aarch64 (were the version for Meson is 0.64). Updating the image to a recent version will fix this problem.

Now, the Windows version is failing because is trying `pip` and not `pip3` or something like that but this version should be fine, installing Meson >= 1.0

subprojects/klib.wrap
```diff
@@ -0,0 +1,8 @@
+[wrap-git]
+url = https://github.com/attractivechaos/klib.git
+revision = cdb7e9236dc47abf8da7ebd702cc6f7f21f0c502
+directory = klib
+patch_directory = klib
+
+[provide]
+klib = klib_dep
```
a1lu, line 8:
```suggestion
klib = klib_dep

```

Newline at EOF

MrAnno (approved):
Nice! Thank you.

PR #536, opened 2024-05-04
Use `nix-shell` instead of of `nix-env -iA`

Sigmanificient:
`nix-env` for installing packages is a bad habit for nix/NixOS users and it is not recommended from the community as it doesn't mix well user configuration.nix / flake and/or home-manager. It also alters the system rather than providing a temporary package.

Since this is within the `download` section, it would make sense to use `nix-shell criterion` which will create a sub-shell with environment variable set to have criterion available.

I would have personally used `nix shell nixpkgs#criterion`, but the `nix` command is still an experimental feature, and isn't enabled by default

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ the user would have with other frameworks:
 **Gentoo** | `emerge -a dev-libs/criterion`
 **Arch Linux** ([AUR][aur]) | `pacaur -S criterion`
 **macOS** | `brew install criterion`
-**Nix** | `nix-env -iA nixpkgs.criterion`
+**Nix** | `nix-shell -p criterion`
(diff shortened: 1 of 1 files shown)


MrAnno (approved):
Thank you :)

I have a very limited understanding of how Nix and NixOS operates, but I've grasped that everything there is based on declarative configuration.

I'm just wondering: I can also see everywhere that it's not recommended to use `nix-env` for installing packages, but the official website does that:

https://nixos.org/guides/how-nix-works/

Is this page outdated or there are valid use-cases for `nix-env -i`?

MrAnno:
(I'll fix the CI soon.)

PR #537, opened 2024-05-05
Replace debugbreak repo to MrAnno's fork

Sigmanificient:
Keep readme sync with `subprojects/debugbreak.wrap` file, which has it's URL set to MrAnno's fork.


MrAnno (approved):
I really don't like spreading forks instead of contributing back to upstream, but I haven't been able to reach the author for long months.

Thanks.

PR #547, opened 2024-07-02
docs: Fix references to JSON/XML output format

jonatanschroeder:
JSON and XML outputs were using the TAP example as output. This PR fixes the reference to the appropriate example.


MrAnno (approved):
Thank you.

Issue #521, opened 2024-02-21
assertion for string should take const char * instead of char *


2LeoCode:
when doing for example:

```
const char * foo = "foo";

cr_assert(eq(str, foo, "foo"));
```
we get a compiler warning because eq(str, ...) expects a char * as second argument, so we loose the const qualifier for foo, but it should take a const char * instead

jakeru:
Right! This issue is similar to #512.

As a workaround you may use `cr_expect_str_eq()` for strings  and `cr_expect_arr_eq` to compare arrays. Both are part of the "old" API.

PR #554, opened 2024-08-18
Allow to assert on const data

quarthex:
This may resolve issue #521.

### DIFF
--- a/include/criterion/internal/assert/op.h
+++ b/include/criterion/internal/assert/op.h
@@ -205,7 +205,7 @@
 #define CRI_ASSERT_OP_TYPE_TAGGED(Type, Var, Name, MkNode, Val)         Type
 #define CRI_ASSERT_OP_TYPE_SINGLE(Type, Var, Name, MkNode, Val)         Type
 
-#define CRI_ASSERT_OP_ARRTYPE_TAGGED(Type, Var, Name, MkNode, Val)      Type *
+#define CRI_ASSERT_OP_ARRTYPE_TAGGED(Type, Var, Name, MkNode, Val)      Type const *
 #define CRI_ASSERT_OP_ARRTYPE_SINGLE(Type, Var, Name, MkNode, Val)      Type
 
 #define CRI_ASSERT_OPKIND(Kind, Type, Var, Name, MkNode, Val)           Kind
--- a/include/criterion/internal/assert/tag.h
+++ b/include/criterion/internal/assert/tag.h
@@ -36,6 +36,8 @@
 
 #define CRI_ASSERT_TYPE_TAG(Tag)           CR_EXPAND(CRI_ASSERT_TYPE_TAG_(Tag))
 #define CRI_ASSERT_TYPE_TAG_(Tag)          CRI_IF_DEFINED_NODEFER(CRI_ASSERT_TEST_TAG_ ## Tag, CR_VA_HEAD(CRI_ASSERT_TYPE_TAG_ ## Tag), , CRI_ASSERT_TYPE_TAG_UNKNOWN, (Tag))
+#define CRI_ASSERT_MUT_TYPE_TAG(Tag)       CR_EXPAND(CRI_ASSERT_MUT_TYPE_TAG_(Tag))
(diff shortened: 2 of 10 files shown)


quarthex:
Ha ha, everything goes soooo wrong. 😑

[EDIT] was talking about the pipeline

MrAnno:
Sorry for the huge delay.

Thank you for this PR.

I think we should enable `-Werror` only in the CI so that users can build Criterion with less difficulty on exotic platforms.

quarthex:
Hello @MrAnno,

I was trying to merge/resolve the bleeding branch into my PR, but my `uncrustify` version seems more recent than yours.
I use  `Uncrustify-0.83.0_f`.
Which version do you use?

quarthex:
Here’s my output, when calling `uncrustify` by hand:

```
.uncrustify.cfg:48: option 'nl_func_var_def_blk' is deprecated; it has been replaced by 'nl_var_def_blk_end_func_top'.
You can also use 'nl_var_def_blk_end' for additional functionality.

.uncrustify.cfg:104: option 'sp_balance_nested_parens' never works; it has been replaced by 'sp_paren_paren'.

.uncrustify.cfg:121: option 'sp_before_tr_emb_cmt' is deprecated; did you want to use 'sp_before_tr_cmt' instead?

.uncrustify.cfg:167: option 'sp_num_before_tr_emb_cmt' is deprecated; did you want to use 'sp_num_before_tr_cmt' instead?

Option<UNUM>: at .uncrustify.cfg:202: Expected unsigned number , for 'align_nl_cont'; got 'true'
.uncrustify.cfg:230: option 'pp_space' is deprecated; it has been replaced by 'pp_space_after'.
```

quarthex:
Another question: is it still OK for you to enable a stricter build in the CI context? Maybe by detecting an environment variable set by GitHub Action, idk?

MrAnno:
@quarthex Don't worry about that :) To be honest, I stopped using uncrustify and forgot about it.
It would probably be worth reviving code formatting one way or another.

----

Sorry for ignoring your PR for so long. I've been thinking about const correctness a lot, but I still don't understand Criterion's little tricks well enough to propose a fix that would work in all scenarios. There were moments when I thought it wasn't even possible.

Let me share the problems I found, you may have a better understanding of them.

C only allows adding qualifiers at the "outermost level". `char *arr[]` usually turns into `char **`, and `char **` cannot convert to `const char *const *` because the pointee types `char *` and `const char *` are incompatible. 

Your PR fixes a few important cases, but breaks other complex cases. These are all around the "pointer to pointer" topic.

Here are a few test cases that I would expect to compile without warnings (I'm not 100% sure all of them make sense, let me know if you'd drop any of them):
[bug521.c](https://gist.github.com/MrAnno/a3e2c36699250d8fde7aee2dea1a224e)

We should also be careful about the following:
- user-types and their constness (`cr_user_<T>_eq`,  `cr_user_<T>_tostr`)
- C++ and untagged `void *`

MrAnno:
> Another question: is it still OK for you to enable a stricter build in the CI context? Maybe by detecting an environment variable set by GitHub Action, idk?

Sure, we can set CFLAGS or anything needed when calling the `meson setup` in the CI, or we can add a strict/pedantic flag directly.

quarthex:
- [x] test `str2`: It is the same thing as `str`. The `static` does nothing.
- [x] test `ptr_to_const_pointer`: no warning.
- [ ] test `str_array_of_mutable_string`: tries to put a `char *[]` into a `const char *const *`.
- [ ] test `str_array_of_const_pointers_to_mutable_strings`: tries to put a `char *const[]` into a `const char *const *`
- [x] test `str_array_of_const_strings`: works at intend.
- [ ] test `str_array_mutable_actual_const_expected`:  a mix of `str_array_of_mutable_strings` and `str_array_of_const_strings`.
- [ ] test `wcs_array_of_mutable_stings`: easy to fix once `str`/`str[]` tags are addressed.
- [ ] test `ptr_array_of_void_pointers`: tries to put a `void *[]` into a `const void *const *` + call `cr_user_ptr_tostr()` with `const void *const *` instead of `const void **`.
- [ ] test `ptr_array_of_const_void_pointers`: call `cr_user_ptr_tostr()` with `const void *const *` instead of `const void **`.
- [x] test `user_type_array_with_mutable_callbacks`: call `cr_user_<type>_tostr()` with `const struct *` instead of `struct *`. IMO, `cr_user_<type>_tostr()` and `cr_user_<type>_eq()` should actually not mutate the value.
- [x] test `const_user_type_array_with_const_callbacks`: works fine.

That *a lot* of work. :sweat:

MrAnno:
Yeah :( I started doing it a year ago, but I got stuck.

MrAnno:
I would even be happy with a partial fix, but as far as I remember, I couldn't fix the simple const cases without breaking other things.

PR #557, opened 2024-12-27
Update FindCriterion.cmake module

petk:
- CMake minimum required version bumped to 3.14 otherwise at current versions there are warnings emitted. And for the `if(IN_LIST)` usage.
- Duplicate CMake find module removed for simplicity
- Added the `Criterion::Criterion` imported library to CMake find module so it can be linked to targets and provide usage requirements
- Added pkg-config integration to the CMake find module


dev/cmake/cmake/FindCriterion.cmake
```diff
@@ -1,27 +1,100 @@
...
+    add_library(Criterion::Criterion UNKNOWN IMPORTED)
+    set_target_properties(
+      Criterion::Criterion
+      PROPERTIES IMPORTED_LOCATION "${Criterion_LIBRARY}"
+    )
+  else()
+    add_library(Criterion::Criterion INTERFACE IMPORTED)
+    set_target_properties(
+      Criterion::Criterion
+      PROPERTIES IMPORTED_LIBNAME "${Criterion_LIBRARY}"
```
MrAnno, line 92: I think `INTERFACE_INCLUDE_DIRECTORIES` should be set too in both cases.

MrAnno (commented):
Thank you very much for this (and sorry for the huge delay).

I am glad to see this, this is exactly how modern CMake should look like.

dev/cmake/cmake/FindCriterion.cmake
petk: Hello, this one is also set at the end of the module in this file.

MrAnno: Indeed. Sorry for the oversight.

MrAnno (approved):

Issue #567, opened 2025-05-07
Clarification on the accepted types for theory data points


sebascert:
I ran into this issue when attempting to define some data points as a struct of
total size 16, upon doing this I got this error and my theory crashing for no
apparent reason:

`boxfort-worker: ../src/core/theories.c:72: cri_theory_push_arg: Assertion 'type' failed.`

When I inspected the source code referenced by the error, I realized that this
is because a data point must be of size no longer than `sizeof(uint64_t)`, I
checked the docs and saw no explicit rule on this.

```c
void cri_theory_push_arg(struct cri_theory_ctx *ctx, bool is_float, size_t size, void *ptr)
{
    ffi_type *float_ffi_types[] = {
        [sizeof (float)] = &ffi_type_float,
        [sizeof (double)] = &ffi_type_double,
        [sizeof (long double)] = &ffi_type_longdouble,
    };

    /* supported sizes are capped at sizeof(uint64_t) */
    ffi_type *ffi_types[] = {
        [sizeof (uint8_t)] = &ffi_type_uint8,
        [sizeof (uint16_t)] = &ffi_type_uint16,
        [sizeof (uint32_t)] = &ffi_type_uint32,
        [sizeof (uint64_t)] = &ffi_type_uint64,
    };

    ffi_type *type = is_float ? float_ffi_types[size] : ffi_types[size];
    assert(type); /* line 72 */
    cri_append_arg(ctx, type, ptr);
}
```

It would be nice if it was mentioned in the docs, and if the error would be less cryptic.

PR #568, opened 2025-05-07
docs: Add clarification on allowed types for theory data points.

sebascert:
This PR adds a clarification into the docs for theory data points, specifying
that the size of the given type for a data point must not exceed
sizeof(uint64_t), as Criterion only supports types with sizes up to that limit.

Intended to solve issue #567


MrAnno (approved):
Thank you

PR #588, opened 2025-12-17
Mark ReportHook & ParameterizedTest as used

jamesjer:
With criterion 2.4.3, I tried running the cram tests on an x86_64 Fedora 43 machine.  The parameterized and report tests failed due to missing output.  In both cases, the linker had discarded the test code while building the executable.  This PR adds `used` attributes to prevent that from happening.  With this change, all of the cram tests pass on my machine.


MrAnno:
Hi @jamesjer,

Thanks for your PR!

I'll try to reproduce the issue soon.

Doesn't this mean that we might be missing additional symbols that are not directly mentioned in the executable?
If so, using Meson's `link_whole` (`-Wl,--whole-archive`) flags may solve this generally.

PR #596, opened 2026-05-24
cli: add --default-timeout option for tests with no explicit timeout

### LINKED ISSUE #279: Criterion hangs (2.3.3)
Hi,

So I was using Criterion 2.3.2-1 (from the AUR) and had some issues with clang++ and cr_assert_throw.
Seeing on of the issues solved in the 2.3.3, I recompiled Criterion to use it.
Now, I have previously successful tests hanging.
The timeout option does not seems to change anything.

Test file:
```C
static bool function_loop()
{
    for(;;);
    return true;
}

Test(simple, test3)
{
    cr_assert(function_loop());
}
```

compilation:
```bash
g++ -O0 -std=c++11 -Wall -Wextra   -lcriterion   test.cc   -o test
```
or
```bash
clang++ -O0 -std=c++11 -Wall -Wextra   -lcriterion   test.cc   -o test
```
version check:
```
$ ./test --version
Tests compiled with Criterion v2.3.3
$ ./test --timeout 2
...
...
```

expected: test failing with a timeout.
got: test looping, 1 core at ~100%

Misc Information:
- Archlinux, 4.19.4, up to date.
- clang, 7.0.0.0 RELEASE_700
- g++ 8.2.1

All the best,

Nathan

PoQuatre:
This PR introduces a new `--default-timeout` CLI option (and its corresponding `criterion_options.default_timeout` field) that applies a fallback timeout to any test that has no timeout set, neither directly on the test nor inherited from its suite.

The existing `--timeout` option acts as a hard cap: it overrides every test's timeout downward if the test's own timeout exceeds the cap. `--default-timeout` fills the complementary gap: it only kicks in when no timeout would otherwise apply, leaving explicitly-timed tests completely untouched.

This would also somewhat resolve #279

## Timeout resolution order (after this PR)

1. Suite timeout (if set)
2. Per-test `.timeout` (overrides suite)
3. `--default-timeout` applied only when the result of steps 1–2 is `<= 0`
4. `--timeout` cap applied only when the result so far exceeds the cap

## Code style notes

Uncrustify was run over the touched files. Two issues came up:

- **Pre-existing errors in untouched files.** Uncrustify reported style violations in files not modified by this PR. Those were left alone to keep the diff focused and avoid muddying the review with unrelated formatting churn.

- **Backslash alignment in the `USAGE` macro (`src/entry/params.c`).** Uncrustify suggested de-aligning the continuation backslashes added for the new `--default-timeout` line. The rest of the macro uses aligned backslashes as an intentional style choice, so the new lines were written to match the surrounding code rather than follow the tool's suggestion.

### Location of the new tests

The cram test (`test/cram/timeout.t`) references `default-timeout.c.bin` by name, which is only built when the full test suite is configured. It is unclear whether this file belongs under `test/cram/` (which typically contains self-contained script tests) or whether it would be better placed alongside the full-test fixtures it depends on, or driven from `test/full/meson.build` directly.

### Help message clarity for `--timeout` vs `--default-timeout`

The current help strings are:

```
--timeout [TIMEOUT]: set a timeout (in seconds) for all tests
--default-timeout [TIMEOUT]: set a default timeout (in seconds) for tests with no explicit timeout
```

These may not make the semantic difference obvious enough to end users, since "for all tests" could be read as "applies universally" without conveying that it acts as a cap. A clearer phrasing might be something like:

```
--timeout [TIMEOUT]: cap the timeout (in seconds) for all tests, overriding any per-test value that exceeds it
--default-timeout [TIMEOUT]: set a fallback timeout (in seconds) applied only to tests with no timeout set
```

### DIFF
--- a/src/core/runner_coroutine.c
+++ b/src/core/runner_coroutine.c
@@ -344,6 +344,8 @@ static bxf_instance *run_test(struct run_next_context *ctx,
         timeout = ctx->test->data->timeout;
     if (criterion_options.timeout > 0 && timeout > criterion_options.timeout)
         timeout = criterion_options.timeout;
+    if (timeout <= 0 && criterion_options.default_timeout > 0)
+        timeout = criterion_options.default_timeout;
 
     sp.iquotas.runtime = timeout;
 
--- a/doc/env.rst
+++ b/doc/env.rst
@@ -20,7 +20,10 @@ Command line arguments
   a number of jobs ideal for your hardware configuration.
 * ``--filter [PATTERN]``: Run tests whose string identifier matches
   the given shell wildcard pattern (see dedicated section below). (\*nix only)
-* ``--timeout [TIMEOUT]``: Set a timeout (in seconds) for all tests
+* ``--timeout [TIMEOUT]``: Cap the timeout (in seconds) for all tests,
+  overriding any per-test value that exceeds it.
+* ``--default-timeout [TIMEOUT]``: Set a fallback timeout (in seconds)
+  applied only to tests with no timeout set.
 * ``--debug[=debugger]``: Run tests with a debugging server attached.
   ``debugger`` can be 'gdb', 'lldb', or 'windbg' (windows only).
 * ``--debug-transport [TRANSPORT]``: Make the debugging server use the
--- a/include/criterion/options.h
+++ b/include/criterion/options.h
@@ -184,6 +184,15 @@ struct criterion_options {
     double timeout;
 
     /**
+     *  A default timeout applied to each test that has no timeout set, in seconds.
+     *
+     *  If the value is non-positive, no default is applied.
+     *
+     *  default: 0
+     */
+    double default_timeout;
+
+    /**
      * Fully report statistics from test workers, including those that are
      * not reported by default (like passing assertions).
      *
--- a/src/entry/params.c
+++ b/src/entry/params.c
@@ -70,8 +70,12 @@
     "name of the source file on a failure\n"                \
     "    --filter [PATTERN]: run tests matching the "       \
     "given pattern\n"                                       \
-    "    --timeout [TIMEOUT]: set a timeout (in seconds) "  \
-    "for all tests\n"                                       \
+    "    --timeout [TIMEOUT]: cap the timeout (in "         \
+    "seconds) for all tests, overriding any per-test "      \
+    "value that exceeds it\n"                               \
+    "    --default-timeout [TIMEOUT]: set a fallback "      \
+    "timeout (in seconds) applied only to tests with "      \
+    "no timeout set\n"                                      \
     "    --tap[=FILE]: writes TAP report in FILE "          \
     "(no file or \"-\" means stderr)\n"                     \
     "    --xml[=FILE]: writes XML report in FILE "          \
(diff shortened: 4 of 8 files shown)


MrAnno (approved):
Thank you for the PR and the meticulous work :)

> Timeout resolution order

It sounds perfect, it's also backward-compatible.

> Code style notes

Thanks again for your precision, it looks perfect to me.
Yeah, the format Uncrustify wants to produce is far from what we can see on `bleeding`.

> Location of the new tests

Timing-based tests aren't the most warmly welcomed creatures in test suites.

`sample / timeout.c` is already in the test suite, but it doesn't validate much, it's more like a smoke test.

I think the location of `default-timeout.c` is right. What's worth considering is how many times we want to run these tests and for what purpose.
`default-timeout` currently runs three times: once as part of the "full" test suite, and twice again as cram-validated tests.

I think it would be worth not registering `default-timeout` as a "full" test and letting cram validate it instead. Or if you prefer keeping it in both suites, we can remove the `timeout` sample from the "full" test suite instead (running 2 `--always-succeed` smoke tests around timeouts is not really useful).

> Help message clarity for --timeout vs --default-timeout

I like your suggestion, it clarifies everything discussed in #279. Would you like to add it to this PR?

PoQuatre:
Thank you for the review and the kind words!

> I think it would be worth not registering `default-timeout` as a "full" test and letting cram validate it instead.

I'll go with this option, removing the `test()` call for `default-timeout` from `test/full/meson.build` and letting the cram tests be the sole validators. The binary still needs to be built (the cram suite references it by name), so I'll keep the `executable()` definition and just drop the `test()` entry. That leaves `sample / timeout.c` as the one `--always-succeed` smoke test for timeouts, and `timeout.t` as the real validation.

> Would you like to add it to this PR?

Yes, happy to include the updated help strings in this PR. I'll update both `--timeout` and `--default-timeout` to the phrasing from the PR description.

src/core/runner_coroutine.c
```diff
@@ -342,6 +342,8 @@ static bxf_instance *run_test(struct run_next_context *ctx,
         timeout = ctx->suite_set->suite.data->timeout;
     if (ctx->test->data->timeout > 0)
         timeout = ctx->test->data->timeout;
+    if (timeout <= 0 && criterion_options.default_timeout > 0)
+        timeout = criterion_options.default_timeout;
```
MrAnno: Shouldn't we move this snippet after the `criterion_options.timeout` check?
That way, it `--default-timeout` wouldn't interfere with `--timeout`.

PoQuatre: Honestly, I don't have a strong opinion either way. Both orderings are defensible:

- **`default-timeout` first, then `timeout` caps it**: `--default-timeout` only fills in when no per-test timeout is set, and `--timeout` still acts as a hard ceiling on top of whatever value ended up being used, whether it came from the test itself or from `--default-timeout`.
- **`timeout` check first, then `default-timeout` fills in**: makes it so `--default-timeout` never applies when `--timeout` is already set, which reads as a stricter separation between the two flags.

I don't really care which one we go with, feel free to pick whichever feels more correct to you and I'll update the code accordingly.

MrAnno: The second one sounds closer to the new description of the timeout option, I'd go with that.

PoQuatre:
I just made the change you suggested, and switched the order in which the `--timeout` and `--default-timeout` are checked and applied.

MrAnno (approved):
Thanks!

PR #597, opened 2026-06-12
alloc: do not use deprecated std::allocator types

MrAnno:
std::allocator is deprecated in C++17, and has been removed from C++20.


MrAnno:
Cirrus CI has disappeared...

https://cirruslabs.org/

I'll look for alternatives to test on FreeBSD and try to fix the Windows job.

